PR #21, opened 2014-10-19
Define and export setitimer, getitimer.

avieth:
~~Export `ualarm` so that we can have microsecond resolution for our `SIGALRM`s.~~

Formerly "Define and export ualarm". That has been dropped in favour of `setitimer` and `getitimer`.

### DIFF
--- a/System/Posix/Time.hsc
+++ b/System/Posix/Time.hsc
@@ -18,9 +18,24 @@
 -----------------------------------------------------------------------------
 
 module System.Posix.Time (
-        epochTime,
-        -- ToDo: lots more from sys/time.h
-        -- how much already supported by System.Time?
+
+      epochTime
+
+    #ifdef HAVE_SETITIMER
+    , ITimerType(..)
+
+    , TimeVal(TimeVal)
+    , timeValSeconds
+    , timeValMicroseconds
+
+    , ITimerVal(ITimerVal)
+    , itimerValInterval
+    , itimerValValue
+
+    , setitimer
+    , getitimer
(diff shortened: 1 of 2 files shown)


avieth:
ualarm dropped in favour of setitimer. Requesting another review.

System/Posix/Time.hsc
```diff
@@ -39,3 +54,92 @@ epochTime = throwErrnoIfMinus1 "epochTime" (c_time nullPtr)
 
 foreign import ccall unsafe "__hsunix_time"
   c_time :: Ptr CTime -> IO CTime
+
+#ifdef HAVE_SETITIMER
```
hs-viktor, line 58: Which still supported POSIX platforms don't have setitimer?  Likely the `#if` guards are long redundant.

System/Posix/Time.hsc
```diff
@@ -39,3 +54,92 @@ epochTime = throwErrnoIfMinus1 "epochTime" (c_time nullPtr)
...
+
+-- | Haskell expression of struct timeval: seconds and microseconds.
+data TimeVal = TimeVal {
+    tv_sec :: Int
```
hs-viktor, line 62: Does GHC still support any 32-bit CPUs with a 64-bit time_t?  Perhaps unofficially on some NetBSD system?  If so, `Int` may be large enough to correctly represent a C `time_t`.  The correct type here may be `CTime` from `Foreign.C.Types`

hs-viktor:
Rebased in #177. Closing this PR.

PR #41, opened 2015-03-02
Fix for when cross-compiling with clang to aarch64.

sseefried:
Just turns off a warning that caused problems when cross compiling from Mac OS X to aarch64
using Xcode 6.1's cross compiler.


sseefried:
This flag just turns off a warning so I don't think it's going to hurt anything.

rwbarton:
Can you test whether you can actually link an executable that uses fileSynchroniseDataOnly after applying this patch?

argiopetech:
Solved by #42.

PR #46, opened 2015-09-25
Add BlockSize and BlocksCount fields for FileStatus

guibou:
(This is a request for discussion, there is, at least, one point which must be handled before merging this code)

The system stat function returns blocksize and blocks count for the file, which are not handled by current implementation. I added the accessor for both fields.
## Discussion

I think the new types aliases:

``` haskell
type BlockSize = CSsize
type BlocksCount = CSsize
```

Should be placed somewhere else (in base ?) and I'm not really sure that's the correct types.

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -57,7 +57,7 @@ module System.Posix.Files (
     getFileStatus, getFdStatus, getSymbolicLinkStatus,
     -- ** Querying file status
     deviceID, fileID, fileMode, linkCount, fileOwner, fileGroup,
-    specialDeviceID, fileSize, accessTime, modificationTime,
+    specialDeviceID, fileSize, blockSize, blocksCount, accessTime, modificationTime,
     statusChangeTime,
     accessTimeHiRes, modificationTimeHiRes, statusChangeTimeHiRes,
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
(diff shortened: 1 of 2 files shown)


argiopetech:
Thanks for your work on this, guibou.

Regarding the aliases, those would go in `System.Posix.Types`. That will require a pull request against `base`. I'm honestly not sure what the historical context is for those definitions to be in base rather than `unix`, but I imagine there's a good reason for it. We'll need to bump the lower bound on the `base` version dependency to match the upcoming base release in order to ensure that type is available.

argiopetech:
Oh, and you are correct about the types needing to be changed. `CSsize` is equivalent to `size_t`, but the current POSIX spec specifies `blkcount_t` and `blksize_t` for `st_blksize` and `st_blocks` (respectively). Those will need to be added as new types using the `INTEGRAL_TYPE` or `INTEGRAL_TYPE_WITH_CTYPE` macros.

DanielG:
I just created a trac ticked for this and patches to both base and unix are forthcoming: https://ghc.haskell.org/trac/ghc/ticket/12795

guibou:
@argiopetech Thanks to @DanielG changes to `base`, I can now update this PR without the ugly type aliases.

(Yes, two years had passed ;)

DanielG:
Probably should have mentioned this before but I also have a PR open for this: https://github.com/haskell/unix/pull/78. It handles platforms where `st_blksize` doesn't exist which this PR doesn't.

guibou:
@DanielG good point. I'm closing in favor of #78

PR #53, opened 2015-12-10
Use -Werror=implicit-function-declaration when checking for fdatasync

### LINKED ISSUE #52: configure is broken
Commit `b06446edd4753f964a46d27ddae864fad97adc13` has changed the semantics of `configure` that it now fails to properly detect if `fdatasync(2)` is available on the given platform.  For more details, please see [ticket #11137 of GHC](https://ghc.haskell.org/trac/ghc/ticket/11137).

thomie:
This fixes #52.

Testcase:
  * in configure.ac, change `fdatasync(4)` to `foo(4)`
  * autoconf && ./configure
  * notice that configure now reports: `checking for fdatasync... no`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,10 +76,13 @@ AC_CHECK_FUNCS([fsync])
 
 # A more comprehensive check that fdatasync exits
 # Necessary for platforms that have fdatasync in headers but have no
-# implementation
+# implementation. Use -Werror=implicit-function-declaration for platforms
+# that don't have fdatasync at all.
 dnl Originally provided by user copiousfreetime for the beanstalkd project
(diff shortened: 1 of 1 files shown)


hvr:
1. Tbh, I really dislike to abuse `-Werror=implicit-function-declaration`  for that
2. There's this instead for testing prototypes: https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Generic-Declarations.html
3. We should probably rewrite the `fdatasync`-autoconf test in terms of `AC_CHECK_FUNCS` and `AC_CHECKS_DECLS`

thomie:
Agreed.

PR #54, opened 2016-01-28
Add functions that are missing on Android

rwbarton:


rwbarton:
I have built the unix package for Android with these changes and verified that the functions involved appear to work as expected (`drainOutput` does not appear to do anything and `getControllingTerminalName` returns `"/dev/tty"`, which even exists and acts as normal).

cbits/ctermid.c
```diff
@@ -0,0 +1,21 @@
+/* Provide ctermid() if it is missing (as on some Android versions). */
+
+#include "HsUnixConfig.h"
+#ifndef HAVE_CTERMID
```
hvr: Fwiw, I consider the use of `#ifdef`s for boolean macros bad style; I know that the `unix` package currently still has a lot of `#ifdef HAVE_*` around, but I'm considering replacing those at some point

cbits/ctermid.c
```diff
@@ -0,0 +1,21 @@
...
+
+#include "ghcplatform.h"
+
+#if defined(linux_HOST_OS) || defined(linux_android_HOST_OS)
```
hvr: why are we testing for a specific OS, rather than follow Autoconf's idiom of testing for features?

cbits/ctermid.c
```diff
@@ -0,0 +1,21 @@
...
+
+char *ctermid(char *s)
+{
+  static const char devtty[] = "/dev/tty";
+  if (s)
+    return strcpy(s, devtty);
+  else
+    return (char *)devtty;
+}
+#endif /* linux or linux_android */
```
hvr: what happens for non-linux platforms lacking ctermid?

cbits/tcdrain.c
```diff
@@ -0,0 +1,21 @@
...
+
+#include "HsUnixConfig.h"
+#ifndef HAVE_TCDRAIN
+
+#include "ghcplatform.h"
+
+#ifdef linux_android_HOST_OS
+#include <sys/ioctl.h>
+
+int tcdrain(int fd)
```
hvr: I don't like this, as we'd be defining POSIX defined `tcdrain`/`ctermid` identifiers into a global namespace for which we have no ownership. We shouldn't need any cbits at all for providing a graceful fallback implementation.

hvr:
In conclusion, I suggest a more principled way to address this issue.

cbits/ctermid.c
rwbarton: They are out of luck, just like they were before this patch. I don't know what the correct terminal name is on such systems. If others encounter the same issue on other systems, they can add an implementation to this file with the correct behavior for their system.

cbits/ctermid.c
rwbarton: Because the implementation is only necessarily correct on Linux systems, where `/dev/tty` is the name of the controlling terminal. On other exotic systems, perhaps the name is something else. What feature would you like to detect with Autoconf for this, while remaining friendly to cross-compilation?

cbits/ctermid.c
rwbarton: Feel free to make that change at some point, including the occurrences in this file.

cbits/tcdrain.c
rwbarton: We could rename the function that we import, like we do with `__hsunix_execvpe` (except identifiers containing a double underscore are also reserved, oops!) In principle that would matter if we linked against another library that also noticed that it was missing the same function and provided its own implementation under the same name.

I don't think pushing all the platform-specific logic (including a separate foreign import of `ioctl()`) into Haskell is a cleaner approach. Particularly in the case of `ctermid()` which is imported from two modules which are almost entirely duplicates of each other.

rwbarton:
I can change the function names to `hsunix_tcdrain` and `hsunix_ctermid`.

Issue #23, opened 2014-10-22
Rethink API wrt to operations not available on all platforms


hvr:
Some time ago, based on the discussion in

  http://thread.gmane.org/gmane.comp.lang.haskell.libraries/20865/focus=20936

@JohnLato prepared a patch at

  https://github.com/JohnLato/packages-unix/commit/a1d905c4cce27c1efab834f1afaaa0d7f9bf1643

that's risking to be forgotten.

PR #59, opened 2016-05-01
Add support for more OpenFileFlags and slightly refactor 'openFd'

### LINKED ISSUE #57: Missing O_NOFOLLOW
This is really annoying. I don't even understand why this library attempts to pre-select which flags I can pass in or not. That's not how a low-level library should be designed.

Since this library is about POSIX stuff, you should simply allow **all** POSIX flags, see http://www.unix.com/man-page/posix/3p/open/

> O_NOFOLLOW    If path names a symbolic link, fail and set errno to [ELOOP].

### LINKED ISSUE #58: "exclusive = True" may lead to undefined behavior
It's not very well documented, because `O_CREAT` is not passed like the other flags via `OpenFileFlags`, but implicitly via `Maybe FileMode`.

The [manpage](http://www.unix.com/man-page/posix/3p/open/) says that
> If O_EXCL is set and O_CREAT is not set, the result is undefined.

This library should either make the correspondence of those input arguments clear or make it impossible to use the combination. The following would already be quite an improvement:
```hs
-- |Correspond to some of the int flags from C's fcntl.h.
data OpenFileFlags =
 OpenFileFlags {
    append    :: Bool,           -- ^ O_APPEND
    exclusive :: Bool,           -- ^ O_EXCL, result is undefined if O_CREAT is False
    noctty    :: Bool,           -- ^ O_NOCTTY
    nonBlock  :: Bool,           -- ^ O_NONBLOCK
    trunc     :: Bool,           -- ^ O_TRUNC
    creat     :: Maybe FileMode  -- ^ O_CREAT 
 }
```
And then remove the extraneous `Maybe FileMode` arguments from `open_`.

hasufell:
Fixes #57
Fixes #58
Wrt #6 (only partly fixed)

### DIFF
--- a/System/Posix/IO.hsc
+++ b/System/Posix/IO.hsc
@@ -76,13 +76,12 @@ import System.Posix.Internals ( withFilePath )
 -- for information on how to use the 'FileMode' type.
 openFd :: FilePath
        -> OpenMode
-       -> Maybe FileMode -- ^Just x => creates the file with the given modes, Nothing => the file must exist.
        -> OpenFileFlags
        -> IO Fd
-openFd name how maybe_mode flags = do
-   withFilePath name $ \str -> do
+openFd name how flags =
+   withFilePath name $ \str ->
      throwErrnoPathIfMinus1Retry "openFd" name $
-       open_ str how maybe_mode flags
+       open_ str how flags
 
 -- |Create and open this file in WriteOnly mode.  A special case of
 -- 'openFd'.  See 'System.Posix.Files' for information on how to use
@@ -90,4 +89,4 @@ openFd name how maybe_mode flags = do
 
 createFile :: FilePath -> FileMode -> IO Fd
 createFile name mode
-  = openFd name WriteOnly (Just mode) defaultFileFlags{ trunc=True }
+  = openFd name WriteOnly defaultFileFlags{ trunc=True, creat=(Just mode) }
--- a/System/Posix/IO/ByteString.hsc
+++ b/System/Posix/IO/ByteString.hsc
@@ -76,13 +76,12 @@ import System.Posix.ByteString.FilePath
 -- for information on how to use the 'FileMode' type.
 openFd :: RawFilePath
        -> OpenMode
-       -> Maybe FileMode -- ^Just x => creates the file with the given modes, Nothing => the file must exist.
        -> OpenFileFlags
        -> IO Fd
-openFd name how maybe_mode flags = do
-   withFilePath name $ \str -> do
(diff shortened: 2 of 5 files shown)


hasufell:
Another way could be like the `posix-paths` module does:
https://github.com/JohnLato/posix-paths/blob/master/src/System/Posix/Directory/Foreign.hsc

which then generates those Flags conveniently from `Foreign.hsc` into [Foreign.hs](https://gist.github.com/hasufell/51ffa3492b30e64a859bb83ceadb118a)... this also allows for platform-specific ifdefs.

The `_open` function then would look something like this:
```hs
open_  :: CString
       -> OpenMode
       -> [Flags]
       -> Maybe FileMode
       -> IO Fd
open_ str how optional_flags maybe_mode = do
    fd <- c_open str all_flags mode_w
    return (Fd fd)
  where
    all_flags  = unionFlags $ optional_flags ++ [open_mode] ++ creat

    (creat, mode_w) = case maybe_mode of
                        Nothing -> ([],0)
                        Just x  -> ([oCreat], x)

    open_mode = case how of
                   ReadOnly  -> oRdonly
                   WriteOnly -> oWronly
                   ReadWrite -> oRdwr
```

hasufell:
seems there is zero interest?

cartazio:
Busy time of year for many !

Unrelatedly: for our posix apis should we be using closed sums or slowly
moving over to pattern synonyms of the bit masks / flags?

erikd:
@cartazio I'd prefer closed sums for now.

As for the code, I'm on the fence. Its going to break some stuff. Its a matter of whether its worth it. I'd be curious to see what @hvr thinks.

Pastafarianist:
Can some of the developers look into this? Seems like a very useful PR.

hasufell:
No time to maintain this. Developers are not responsive.

cartazio:
@hvr you never gave feedback on this :(

hvr:
Sorry, totally missed this one. So I see a couple of issues with the current approach; afaik the additional flags (O_NOFOLLOW, O_CLOEXEC, O_DIRECTORY and O_SYNC) were recent additions to the POSIX spec (they weren't yet part of the 2004 edition; they started appearing in the 2016 edition), and I'm not sure about how portable they are, so we may need to account for the possibility they aren't available everywhere.

And this poses the question how to deal when they're not available, and how to signal this to API consumers (c.f. #23) . Another issue is that the representation chosen requires us to make version major bumps whenever we add or remove fields from the `OpenFileFlags` type. I'd prefer a scheme which avoids that.

PR #61, opened 2016-05-10
getEnvDefault: Make the fallback explicit in docs

alexander-b:
Signed-off-by: Alexander Berntsen <alexander@plaimi.net>

### DIFF
--- a/System/Posix/Env/ByteString.hsc
+++ b/System/Posix/Env/ByteString.hsc
@@ -53,8 +53,8 @@ getEnv name = do
      else return Nothing
 
 -- |'getEnvDefault' is a wrapper around 'getEnv' where the
--- programmer can specify a fallback if the variable is not found
--- in the environment.
+-- programmer can specify a fallback as the second argument, which will be
+-- used if the variable is not found in the environment.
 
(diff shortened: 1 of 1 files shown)


System/Posix/Env/ByteString.hsc
```diff
@@ -53,8 +53,8 @@ getEnv name = do
      else return Nothing
 
 -- |'getEnvDefault' is a wrapper around 'getEnv' where the
--- programmer can specify a fallback if the variable is not found
--- in the environment.
+-- programmer can specify a fallback as the second argument, which will be
+-- used if the variable is not found in the environment.
```
hs-viktor, line 57: Would it perhaps be better to instead (or also) add haddock comments for each of the parameters in the signature? Perhaps in that case leaving the original description unchanged.  The first argument can then be clearly indicated as the name of the environment variable, and the second as the fallback value.

hasufell: As the original author is no longer with us I think both is fine.

hs-viktor:
Closing, obsoleted by #176

PR #64, opened 2016-06-02
Add GHC 8.0.1 to test-matrix.

tmcgilchrist:
What are the plans for supporting 8.0?

As is this isn't likely to work given the constraints in the cabal file.


glguy:
`unix-2.7.2.0` is already distributed with GHC 8.0.1. Is that what your question is asking about?

glguy:
It would probably be good for us to have 7.10.2 and 7.10.3 in the matrix as well (there seems to be an existing precedent for having all the minor versions).

Also, why is 8.0.1 added as allowed failure?

glguy:
Resolved in 7207bae8e0a5a0885041728f11f19fde46ed0066

PR #65, opened 2016-06-04
Clarify error message for createLink, createSymbolicLink and rename

### LINKED ISSUE #60: Misleading error messages in `rename`
The `rename` function always mentions the first argument even if the issue is with the second one. For example:

```
rename "file.txt" "directory-which-does-not-exist/file.txt"
```

if `file.txt` exists, but `directory-which-does-not-exist` does not, the error message will say:

```
file.txt: rename: does not exist (No such file or directory)
```

oherrala:
POSIX's ENOENT error doesn't specify if the problem was with source or
destination path. Throw error mentioning both paths instead of only
first one.

Fixes #60

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -229,7 +229,7 @@ createLink :: FilePath -> FilePath -> IO ()
 createLink name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createLink" name1 (c_link s1 s2)
+  throwErrnoIfMinus1_ ("createLink "++name1++" to "++name2) (c_link s1 s2)
 
 -- | @removeLink path@ removes the link named @path@.
 --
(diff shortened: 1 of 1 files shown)


System/Posix/Files.hsc
```diff
@@ -253,7 +253,8 @@ createSymbolicLink :: FilePath -> FilePath -> IO ()
 createSymbolicLink file1 file2 =
   withFilePath file1 $ \s1 ->
   withFilePath file2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createSymbolicLink" file1 (c_symlink s1 s2)
+  throwErrnoIfMinus1_ ("createSymbolicLink "++file1++" to "++file2)
```
harendra-kumar: For a link (symbolic or hard) this message may be confusing. When we say `ln s1 s2`, we are creating s2 which points to s1, is that a `s1 to s2` link or `s2 to s1`?  The [unix manual for symlink](http://man7.org/linux/man-pages/man2/symlink.2.html) uses the terms `linkpath` (s2) and `target` (s1), maybe we can use the same terminology in the error message here. That way it will also be easy to correlate the error message with syscall manual when debugging.

luispedro:
This is still open after >2 years, can it be updated and merged? It's still an issue.

Issue #66, opened 2016-06-23
accessTimeHiRes / modificationTimeHiRes / statusChangeTimeHiRes do not return subsecond precision on OS X


jessekempf:
Hi,

I'm using unix 2.7.2.0 on OS X 10.11.5, Stack is version 1.0.4.3 x86_64, and ghc version 7.10.3. I'm finding that despite the autogenerated `include/HsUnixConfig.h` including
```
#define HAVE_STRUCT_STAT_ST_ATIMESPEC 1
#define HAVE_STRUCT_STAT_ST_CTIMESPEC 1
#define HAVE_STRUCT_STAT_ST_MTIMESPEC 1
```
I'm getting results like:
```
> statusChangeTimeHiRes <$> getFileStatus "/tmp" 
1466637634s
```

When I manually run hsc2hs on a checkout of master, the produced code looks like:
```
accessTimeHiRes (FileStatus stat) =
  unsafePerformIO $ withForeignPtr stat $ \stat_ptr -> do
    sec  <- ((\hsc_ptr -> peekByteOff hsc_ptr 32)) stat_ptr :: IO EpochTime
{-# LINE 292 "System/Posix/Files/Common.hsc" #-}

{-# LINE 296 "System/Posix/Files/Common.hsc" #-}
    nsec <- ((\hsc_ptr -> peekByteOff hsc_ptr 40)) stat_ptr :: IO (Int64)
{-# LINE 297 "System/Posix/Files/Common.hsc" #-}
    let frac = toInteger nsec % 10^(9::Int)

{-# LINE 310 "System/Posix/Files/Common.hsc" #-}
    return $ fromRational $ toRational sec + frat
```

I'm now stuck trying to figure out what to look at next in tracking this problem down. Can you help?

glguy:
Are you storing your files in an HFS+ partition (the default)? HFS+ features a 1-second timestamp resolution!

jessekempf:
@glguy: Well. Yes, as a matter of fact, I am. That would explain it -- for some reason I derped out and didn't write a test in C to figure out what was coming back from `stat(2)`.

Do you think I should add a note about the FS-dependency of timestamp resolution, or is it obvious?

ndmitchell:
The Mac thing surprised me in the past (its amazing an OS still running doesn't have better timestamps), so I think it's worth adding a note.

jessekempf:
[...]

PR #67, opened 2016-06-23
Add documentation on timestamp resolution limitations on HFS+.

jessekempf:


glguy:
I think it's less important what HFS+ has one second resolution and more important that people know that this API isn't going to increase the resolution beyond the source data, which is the filesystem. Every filesystem has its own constraints on timestamp resolution. It's possible that having a table of all the resolutions is a good idea, but I'm not sure that this is the right place for that. Additionally the resolution varies by metadata type. For example on FAT32 (according to wikipedia) we have:

```
2 seconds for last modified time,
10 ms for creation time,
1 day for access date,
2 seconds for deletion time
```

ndmitchell:
@glguy I think it's important people know it's file system dependent. But in practice, on Linux it will usually be high resolution and on Mac it will usually be low resolution. That seems worth knowing to someone reading the API who isn't necessarily an expert in other platform specific differences.

jessekempf:
That's a fair point; I think on #66 what tripped both me and @ndmitchell up was that HFS+ is frequently used for `~` on a popular Unix -- I wouldn't have been surprised were I mounting a FAT32 volume, but I was by OS X's behavior, and it seemed to me plausible that conditional compilation was failing somewhere, somehow (obviously not now, but at the time...). I'm new to Haskell, but not new to autoconf failing in bizarre ways.

So on the one hand, we've got `unix` as a wrapper library around `stat(2)`, `fstat(2)`, `lstat(2)`, etc.; and on the other hand we've got `System.Posix.Files` as the Haskell-y way of interacting with files on Unix -- I think seen from the second perspective keeping a table of resolutions, or at least a reference to a freely available one, is more reasonable than if one looks from the first perspective. Am I off base?

ndmitchell:
I'd go with "Note that the time will only be as precise as the underlying filesystem provides (e.g. on HFS+ Mac OS volumes it will be rounded to the nearest second)".

emilypi:
@jessekempf what's the status of this PR? Is it still in flight (or relevant?)

emilypi:
I'm informed by @jessekempf that it is no longer relevant and he will not be continuing the PR.

emilypi (approved):
It's useful as is, and we can iterate on these changes.

Issue #70, opened 2016-09-05
`readdir_r` is deprecated


erikd:
Building revision 40820da5fb on Debian Testing I get:
```
cbits/HsUnix.c: In function ‘__hscore_readdir’:

cbits/HsUnix.c:64:3: error:
     warning: ‘readdir_r’ is deprecated [-Wdeprecated-declarations]
       res = readdir_r(dirPtr, p, pDirEnt);
       ^~~

In file included from include/HsUnix.h:71:0: error:
    0,
                     from cbits/HsUnix.c:9:

```
 According to the man page:

> This function is deprecated; use readdir(3) instead.
>
> The readdir_r() function was invented as a reentrant version of readdir(3). It reads the next directory entry from the directory stream dirp, and returns it in the caller-allocated buffer pointed to by entry.
>
> It is expected that a future version of POSIX.1 will make readdir_r() obsolete, and require that readdir() be thread-safe when concurrently employed on different directory streams.

Since the implementation already has `#if HAVE_READDIR_R` the fix mostly involves detecting whether `readdir_r` is deprecated or not and setting `HAVE_READDIR_R` appropriately.

erikd:
The opinion some placed is to just drop `readdir_r` all together : https://www.spinics.net/lists/fedora-devel/msg222990.html

hvr:
@erikd fair enough... but is it safe to use `readdir(3)` instead of `readdir_r(3)` on say older Linux LTS distros (or other conservative Unices with very long lifecycles)?

erikd:
@hvr really not sure. The Fedora people are using some ugly CPP to check for Glibc verions:
```
#if !defined(__GLIBC__) || (__GLIBC__ < 2) || ((__GLIBC__ == 2) && (__GLIBC_MINOR__ < 23))

    /* readdir_r is deprecated with newer GLIBC */
    struct dirent entry, *iter = 0;
    while ( (errno = readdir_r( d.d, &entry, &iter )) == 0 && iter ) {
        std::string ename( entry.d_name );
#else
   /* use readdir */
```
Do we want to do something similar?

hvr:
Well, it's just a deprecation warning after all, is it? Do we really have to silence it? If `readdir_r(3)` gets finally removed, our autoconf tests would do the right thing anyway...

erikd:
We need to silence it because without it being silenced, its not possible to build GHC with `-Werror`.

ip1981:
[...]

PR #72, opened 2016-09-06
Don't use readdir_r if deprecated

erikd:
GNU glibc 2.23 and later deprecate `readdir_r` in favour of plain old
`readdir` which in some upcoming POSIX standard is going to required to be
re-entrant.

Eventually we want to drop `readder_r` all together, but want to be
compatible with older unixen which may not have a re-entrant `readdir`.

Solution is to make systems with *known* re-entrant `readir` use that and
use `readdir_r` whereever we have it and don't *know* that `readdir` is
re-entrant.

Closes: https://github.com/haskell/unix/issues/70

### DIFF
--- a/cbits/HsUnix.c
+++ b/cbits/HsUnix.c
@@ -37,12 +37,29 @@ int __hsunix_push_module(int fd, const char *module)
 }
 
 /*
+ * GNU glibc 2.23 and later deprecate `readdir_r` in favour of plain old
+ * `readdir` which in some upcoming POSIX standard is going to required to be
+ * re-entrant.
+ * Eventually we want to drop `readder_r` all together, but want to be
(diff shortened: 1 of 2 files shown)


hvr:
Tbh, I'm not a big fan of having the #ifdef there (I'd rather move it to autoconf-level). But I'm not going to block this PR either if it's blocking GHC dev; I'll move it to Autoconf-phase at some later time though.

erikd:
For the changelog entry, should I start a new section? Should I call it 2.7.2.1?

hvr:
@erikd yes, 2.7.2.1 is fine; if it turns out to be wrong it'll be fixed up shortly before release

erikd:
I did look for a way to do it in autoconf, but didn't find any in a google search.

PR #73, opened 2016-09-08
Fix segfault from inconsistent macro use.

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>


erikd:
Good catch! :100:

PR #77, opened 2016-10-13
Add argument documentation for Env modules

### LINKED ISSUE #76: getEnvDefault: which is which?
`getEnvDefault :: String -> String -> IO String`

“`getEnvDefault` is a wrapper around `getEnv` where the programmer can specify a fallback if the variable is not found in the environment.”

Well, thanks Obama, but what is the variable name and what is the fallback?

glguy:
Fixes #76

### DIFF
--- a/System/Posix/Env.hsc
+++ b/System/Posix/Env.hsc
@@ -50,7 +50,9 @@ import qualified GHC.Foreign as GHC (newCString)
 
 -- |'getEnv' looks up a variable in the environment.
 
-getEnv :: String -> IO (Maybe String)
+getEnv ::
+  String            {- ^ variable name  -} ->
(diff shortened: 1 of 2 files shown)


System/Posix/Env.hsc
```diff
@@ -137,7 +144,7 @@ unsetEnv name = putEnv (name ++ "=")
 -- |'putEnv' function takes an argument of the form @name=value@
 -- and is equivalent to @setEnv(key,value,True{-overwrite-})@.
 
-putEnv :: String -> IO ()
+putEnv :: String {- ^ "key=value" -} -> IO ()
```
hvr: Does haddock interpret those quotes properly or do you need to \-escape them?

PR #78, opened 2016-11-02
Add accessors for st_blocks and st_blksize

DanielG:
AFAIU st_blocks is mandatory in POSIX but st_blksize is part of the XSI option (whatever that means) so it's definition is protected by a feature test macro.

This depends on https://ghc.haskell.org/trac/ghc/ticket/12795.

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -59,6 +59,9 @@ module System.Posix.Files (
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
     isDirectory, isSymbolicLink, isSocket,
 
+    fileBlockSize,
+    fileBlocks,
+
     -- * Creation
     createNamedPipe,
     createDevice,
--- a/System/Posix/Files/Common.hsc
+++ b/System/Posix/Files/Common.hsc
@@ -56,6 +56,9 @@ module System.Posix.Files.Common (
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
     isDirectory, isSymbolicLink, isSocket,
 
+    fileBlockSize,
+    fileBlocks,
+
     -- * Setting file sizes
     setFdSize,
 
@@ -255,6 +258,14 @@ specialDeviceID  :: FileStatus -> DeviceID
 -- | Size of the file in bytes. If this file is a symbolic link the size is
 -- the length of the pathname it contains.
 fileSize         :: FileStatus -> FileOffset
+-- | Number of blocks allocated for this file, in units of
+-- 512-bytes. Returns @Nothing@ if @st_blocks@ is not supported on this
+-- platform.
+fileBlocks       :: FileStatus -> Maybe CBlkCnt
+-- | Gives the preferred block size for efficient filesystem I/O in
+-- bytes. Returns @Nothing@ if @st_blocksize@ is not supported on this
+-- platform.
+fileBlockSize    :: FileStatus -> Maybe CBlkSize
 -- | Time of last access.
 accessTime       :: FileStatus -> EpochTime
 -- | Time of last access in sub-second resolution. Depends on the timestamp resolution of the
@@ -293,6 +304,19 @@ modificationTime (FileStatus stat) =
   unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_mtime)
 statusChangeTime (FileStatus stat) =
   unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_ctime)
+
+#ifdef HAVE_STRUCT_STAT_ST_BLOCKS
+fileBlocks (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blocks)
+#else
+fileBlocks _ = Nothing
+#endif
+#ifdef HAVE_STRUCT_STAT_ST_BLKSIZE
+fileBlockSize (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blksize)
+#else
+fileBlockSize _ = Nothing
+#endif
 
 accessTimeHiRes (FileStatus stat) =
   unsafePerformIO $ withForeignPtr stat $ \stat_ptr -> do
--- a/System/Posix/Files/ByteString.hsc
+++ b/System/Posix/Files/ByteString.hsc
@@ -59,6 +59,9 @@ module System.Posix.Files.ByteString (
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
     isDirectory, isSymbolicLink, isSocket,
 
+    fileBlockSize,
+    fileBlocks,
+
     -- * Creation
     createNamedPipe,
     createDevice,
--- a/configure.ac
+++ b/configure.ac
@@ -66,6 +66,9 @@ AC_CHECK_MEMBERS([struct stat.st_uatime])
 AC_CHECK_MEMBERS([struct stat.st_umtime])
 AC_CHECK_MEMBERS([struct stat.st_uctime])
 
+AC_CHECK_MEMBERS([struct stat.st_blocks])
+AC_CHECK_MEMBERS([struct stat.st_blksize])
+
 AC_CHECK_MEMBER([struct passwd.pw_gecos], [], [AC_DEFINE([HAVE_NO_PASSWD_PW_GECOS],[],[Ignore the pw_gecos member of passwd where it does not exist])], [[#include <pwd.h>]])
 
 # Functions for changing file timestamps


hvr:
Fyi, I need to think about the conditional API part; we're trying hard to avoid conditional exports going forward with `unix`, c.f. #23

DanielG:
How about this then?

hvr:
@DanielG better... otoh, if you know it's conditional, you can also make it into a `Maybe`, that way it can be caught w/o exception handling

System/Posix/Files/Common.hsc
```diff
@@ -279,13 +282,26 @@ specialDeviceID (FileStatus stat) =
...
   unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_ctime)
 
+#ifdef HAVE_STRUCT_STAT_ST_BLKSIZE
+-- | Gives the preferred block size for efficient filesystem I/O in bytes.
+fileBlockSize    :: FileStatus -> CBlkSize
+fileBlockSize (FileStatus stat) =
+  unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blksize)
+#else
+{-# WARNING fileBlockSize "fileBlockSize: not available on this platform" #-}
+fileBlockSize =
```
hvr: btw, this CPP branch is missing its typesig

DanielG: Should be fine now.

DanielG:
@hvr what's the status on this? Any reason this never got merged?

llelf:
> AFAIU st_blocks is mandatory in POSIX

it's not.

Bodigrim:
This is blocked until a) either we drop support of older GHCs, b) or agree our stance on conditional APIs, c) fix backward compatibility in other way.

System/Posix/Files/Common.hsc
```diff
@@ -286,6 +291,22 @@ modificationTime (FileStatus stat) =
 statusChangeTime (FileStatus stat) =
   unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_ctime)
 
+#ifdef HAVE_STRUCT_STAT_ST_BLOCKS
+fileBlocks (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blocks)
+#else
+# WARNING fileBlocks "fileBlocks: not available on this platform" #
```
hs-viktor, line 298: What is the purpose of the WARNING? It will be seen by anyone compiling the package on such a platform, whether they intend to use the function or not, and will not be seen by the caller, where the `Nothing` result adequately conveys the same message in a more targeted manner.  So unless I'm missing something the WARNING is not useful and should be removed.

System/Posix/Files/Common.hsc
```diff
@@ -286,6 +291,22 @@ modificationTime (FileStatus stat) =
...
+fileBlocks _ = Nothing
+#endif
+
+#ifdef HAVE_STRUCT_STAT_ST_BLKSIZE
+fileBlockSize (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blksize)
+#else
+# WARNING fileBlockSize "fileBlockSize: not available on this platform" #
```
hs-viktor, line 306: Ditto.

hs-viktor (changes requested):
Looks largely fine, but unless I'm missing something about who sees the `#WARNING`s about missing optional structure members, those warnings should go.

System/Posix/Files/Common.hsc
```diff
@@ -250,6 +251,10 @@ specialDeviceID  :: FileStatus -> DeviceID
 -- | Size of the file in bytes. If this file is a symbolic link the size is
 -- the length of the pathname it contains.
 fileSize         :: FileStatus -> FileOffset
+-- | Number of block blocks allocated for this file, in units of 512-byte.
+fileBlocks       :: FileStatus -> Maybe CBlkCnt
```
hs-viktor, line 255: The types `CBlkCnt` and `CBlkSize` below don't appear to be defined with `GHC <= 8.0` breaking CI.

hs-viktor: Indeed these were introduced *conditionally* in [GHC 8.2](https://gitlab.haskell.org/ghc/ghc/-/commit/ffc2327070dbb664bdb407a804121eacb2a7c734) and this PR can probably use the corresponding macros to detect their presence.  The new feature can then only be offered if the `stat` structure has the requisite fields (already done) AND the types are available.

hs-viktor:
> This is blocked until a) either we drop support of older GHCs, b) or agree our stance on conditional APIs, c) fix backward compatibility in other way.

I am not sure why "older GHCs" come into play here.  Sure we need at least GHC 8.2 for the types, but since in any case the member fields of "stat" are not universally present, and the PR now returns `Nothing` on platforms where they're missing, it can (and should) also return `Nothing` with older GHCs, so the API will be the same API for all systems, but on some systems the return value will always be `Nothing`.  I think that's acceptable.

In other words, given GHC >= 8.2, and a platform that has these fields in `struct stat` the API will return `Just <value>` for these fields, and otherwise `Nothing`.  I think this PR can move forward mostly as-is, with appropriate CPP conditionals.

System/Posix/Files/Common.hsc
DanielG: Hmm, I think I originally had these as `{-# WARNING #-}` pragmas, which would fire if the function is used, but I agree it doesn't make much sense as a CPP warning.

System/Posix/Files/Common.hsc
DanielG: Ah, right. Not sure how we missed that before. I'll add some more GLASGOW_HASKELL version checking then.

DanielG (commented):
> it can (and should) also return Nothing with older GHCs

No it can't because we can't mention the new types since they don't exist on old GHCs yet.

I think there is simply no (good) way to avoid having a conditional API depending on the GHC version. We can however still offer the same API regardless of the availability of the two struct fields. I think that's acceptable.

DanielG:
Changes:
- Decided to use `MIN_VERSION_base` instead of `GLASGOW_HASKELL` as a more direct check for the CBlk* types.

DanielG:
Changes:
- Fix typo in fileBlocks docs

System/Posix/Files/Common.hsc
hs-viktor: This CPP guard is not sufficient. GHC 8.2 [conditionally](https://gitlab.haskell.org/ghc/ghc/-/commit/ffc2327070dbb664bdb407a804121eacb2a7c734#a619a8fa2d426588e918a7b2676a76fac02c51d7_74_74) defines these types.  So you need to:
```
#include "HsBaseConfig.h"
```
and use the same feature macros.

hs-viktor: Which means that the types might in fact be unsupported on some platforms even with GHC >= 8.2.  However, when that's the case we could define them to be new types for `Void` in that case, and then the functions can return `Maybe X -- X ~ Void`, which can only be instantiated as `Nothing`.

DanielG: I tried that, but that doesn't work because `HsBaseConfig.h` defines a bunch of conflicting defines like PACKAGE_NAME etc. which we also have in `HsUnix.h`.

I'm pretty sure that if the `st_` fields exist the corresponding types must also be defined in the system headers so there shouldn't be any need to check for the types and struct fields separately.

DanielG: > then the functions can return Maybe X -- X ~ Void

I thought about that too, but then the API is conditional depending on platform factors again :)

DanielG: > I'm pretty sure that if the st_ fields exist the corresponding types must also be defined

Looks like these didn't always exist, at least according to https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html

> ### Issue 5
> [...]
> The type of st_blksize is changed from long to blksize_t; the type of st_blocks is changed from long to blkcnt_t.

hs-viktor:
Any interest in rebasing this on master, which now has a more comprehensive CI?
This looks rather close now.  I'd like to see whether it is possible to get sensible behaviour with GHC 8.0, with the functions returning `Nothing :: Maybe SomeIntegralType` for some suitable choice of `SomeIntegralType`.

hs-viktor:
Please rebase, now that #175 is merged.

DanielG:
Sure, thanks for the ping :)

System/Posix/Files.hsc
```diff
@@ -59,6 +59,11 @@ module System.Posix.Files (
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
     isDirectory, isSymbolicLink, isSocket,
 
+#if MIN_VERSION_base(4,10,0)
```
Bodigrim, line 62: You do not need to check this condition any longer, because since #175 `unix.cabal` requires `base-4.10` or newer.

DanielG: Ah! super, thanks :)

DanielG:
Changes:
- Remove #ifdefs for MIN_VERSION_base

Bodigrim (approved):

System/Posix/Files/Common.hsc
```diff
@@ -294,6 +305,19 @@ modificationTime (FileStatus stat) =
 statusChangeTime (FileStatus stat) =
   unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_ctime)
 
+#ifdef HAVE_STRUCT_STAT_ST_BLOCKS
+fileBlocks (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blocks)
```
vdukhovni, line 310: Here, there may be a slight difference in strictness between the above, and:
```suggestion
  unsafePerformIO $ fmap Just $ withForeignPtr stat $ (#peek struct stat, st_blocks)
```
Forcing the suggested variant should (I think) actually perform the IO before returning `Just` the result.  Whereas with the original version we may still have a lazy value.  Which version do you think is the more appropriate?

System/Posix/Files/Common.hsc
```diff
@@ -294,6 +305,19 @@ modificationTime (FileStatus stat) =
...
+#else
+fileBlocks _ = Nothing
+#endif
+#ifdef HAVE_STRUCT_STAT_ST_BLKSIZE
+fileBlockSize (FileStatus stat) =
+  Just $ unsafePerformIO $ withForeignPtr stat $ (#peek struct stat, st_blksize)
```
vdukhovni, line 316: Same idea as above, and again, I'm not actually sure which is more corrrect. @Bodigrim your take?
```suggestion
  unsafePerformIO $ fmap Just $ withForeignPtr stat $ (#peek struct stat, st_blksize)
```

vdukhovni (commented):
Looks ready overall, just curious what others make of my strictness question?

System/Posix/Files/Common.hsc
hs-viktor: The above is likely (intended to be) equivalent to `Just $! unsafePerformIO $ ...`

System/Posix/Files/Common.hsc
Bodigrim: Currently GHC knows constructor in compile time, and can easily substitute it further, eliminating wrapper and associated allocations altogether. I assume that usually a result is almost immediately consumed by `fromMaybe (error "unsupported")` or similar. Swapping `Just` and `unsafePerformIO` will make constructor opaque to GHC, so I'd probably keep it as is.

vdukhovni: Fair enough, I guess. I have an instinctive distaste for pending lazy I/O, but this case pretty harmless, and as you say, in practice likely to be immediately evaluated by the caller.

hs-viktor (approved):
With FreeBSD passing CI after initially timing out in some sort of odd virtualbox error, this appears to be ready. I'm satisfied with the `Just/unsafePerformIO` order rationale by @Bodigrim, so this is ready to go.  So I'm approving and merging this PR.

DanielG:
🎉 This has to be my longest running PR by far, hehe, only took a measly 4.25 years to get merged, thanks everyone :D

vdukhovni:
Congratulations!  Now that we've sorted out the CI, and dropped support for old GHC versions that can make it difficult to adopt new features, it should be possible to address the backlog without similar delays.

PR #84, opened 2017-02-08
Fix error message of `createSymbolicLink`.

nh2:
Consider `ln` (or any other Unix tool):

    $ ln -s file1 file2
    $ ls -l file2
    lrwxrwxrwx 1 niklas niklas 5 Feb  8 03:09 file2 -> file1
    $ ln -s file1 file2
    ln: failed to create symbolic link 'file2': File exists

The file name mentioned in the error ("link2") is the one
that *could not be created*, not the content of the pointer.

`createSymbolicLink` got this wrong so far, it would print

    file1: createSymbolicLink: already exists (File exists)

which is wrong, this file doesn't already exist.

This commit fixes it.

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -253,7 +253,7 @@ createSymbolicLink :: FilePath -> FilePath -> IO ()
 createSymbolicLink file1 file2 =
   withFilePath file1 $ \s1 ->
   withFilePath file2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createSymbolicLink" file1 (c_symlink s1 s2)
+  throwErrnoPathIfMinus1_ "createSymbolicLink" file2 (c_symlink s1 s2)
 
 foreign import ccall unsafe "symlink"
   c_symlink :: CString -> CString -> IO CInt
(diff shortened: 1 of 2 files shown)


glguy:
This seems like a reasonable change. Consulting the man page for symlink(2) on macOS I note that it's only in the case of EFAULT that file1 might be the relevant filename: `[EFAULT]           Path1 or path2 points outside the process's allocated address space.` In all other cases the error message pertains to path2.

nh2:
> points outside the process's allocated address space

@glguy do you know what this means in the context of file systems? I'm trying to trigger this, but don't know how.

nh2:
@glguy I just wanted to implement a check or `EFAULT` to address this, but then noticed that indeed it says `Path1 or path2` so we can't really know which one it is.

So `EFAULT` seems to be unbiases about whether to include the first or second one in the message.

rwbarton:
`EFAULT` means that one of the `char *`s you passed points to a memory address that isn't mapped in the application. For example, it is likely to happen if you pass a totally random `char *` value to `symlink()`. You'll have some trouble triggering it via `createSymbolicLink`; I wouldn't worry about the error message.

nh2:
Ah, that makes sense, thanks for the explanation.

PR #90, opened 2017-03-15
Define _POSIX_VDISABLE, if not defined.

angerman:


bgamari:
Yes, this looks quite familiar from my previous work on Android. Thanks!

PR #92, opened 2017-03-20
Android doesn’t have telldir/seekdir in bionic

### LINKED ISSUE #91: configure sets seekdir/telldir to yes for android cross compilation.
Using gnu gold as a linker, specifically:
GNU gold (binutils-2.25-0666073 2.25.51.20141117) 1.11

compiling and linking:
```
/* Define seekdir to an innocuous variant, in case <limits.h> declares seekdir.
   For example, HP-UX 11i <limits.h> declares gettimeofday.  */
#define seekdir innocuous_seekdir

/* System header to define __stub macros and hopefully few prototypes,
    which can conflict with char seekdir (); below.
    Prefer <limits.h> to <assert.h> if __STDC__ is defined, since
    <limits.h> exists even on freestanding compilers.  */

#ifdef __STDC__
# include <limits.h>
#else
# include <assert.h>
#endif

#undef seekdir

/* Override any GCC internal prototype to avoid an error.
   Use char because int might match the return type of a GCC
   builtin and then its argument prototype would still apply.  */
#ifdef __cplusplus
extern "C"
#endif
char seekdir ();
/* The GNU C library defines this for functions which it implements
    to always fail with ENOSYS.  Some functions are actually named
    something starting with __ and the normal name is an alias.  */
#if defined __stub_seekdir || defined __stub___seekdir
choke me
#endif

int
main ()
{
return seekdir ();
  ;
  return 0;
}
```

does not result in an error, even though `seekdir` is not resolved:

```
$ nm -u a.out
         U __cxa_atexit
         U __libc_init
         U seekdir
```

this then later results in `seekdir` missing when trying to deploy the code.

I was not able to convince gold to complain about the unresolved `seekdir`.

angerman:
The isse here is that while we try to use `AC_CHECK_FUNCS`, this will generate test code
that tries to link an object and check for linking errors.  However  GNU gold at least version
(binutils-2.25-0666073 2.25.51.20141117) 1.11, considers seekdir/telldir as part of the
default libs. Thus we would actually want to pass `-nodefaultlibs` to the linker. Doing so,
would result in erros due to not finding `-ldl`.

Fixes #91

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,8 +37,10 @@ AC_CHECK_FUNCS([nanosleep])
 AC_CHECK_FUNCS([ptsname])
 AC_CHECK_FUNCS([setitimer])
 AC_CHECK_FUNCS([readdir_r])
+
 dnl not available on android so check for it
-AC_CHECK_FUNCS([telldir seekdir])
+AC_CANONICAL_TARGET
+AS_CASE([$target_os],[*-android*],[],[AC_CHECK_FUNCS([telldir seekdir])])
 
 dnl When available, _NSGetEnviron() (defined in <crt_externs.h>) is
 dnl the preferred way to access environ(7)


hvr:
lgtm, thanks

PR #95, opened 2017-05-09
Update config.{guess,sub}

angerman:
Autoconf hasn’t had an update since 2014, and it doesn’t look like it will soon[1]
This updates config.{guess,sub}

It basically does exactly what the config.guess script says:
> It is advised that you download the most up to date version of the config scripts from

It adds support for e.g. `-ios`, which allows to have targets like `aarch64-apple-ios`.

—
[1]: http://lists.gnu.org/archive/html/autoconf/2016-07/msg00017.html

### DIFF
--- a/config.guess
+++ b/config.guess
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2014 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


bgamari:
Sounds reasonable to me. @hvr?

hvr:
sure!

PR #97, opened 2017-06-22
gitignore cabal sandbox and new-build artifacts

gwils:


.gitignore
```diff
@@ -1,6 +1,9 @@
 # Specific generated files
 GNUmakefile
 autom4te.cache/
+.cabal-sandbox/
+cabal.project.local
+cabal.sandbox.config
```
hvr: while at it, the new `ghc.environment.*` should also be ignored

gwils: Ok, done.

PR #101, opened 2017-09-15
Bump upper bound on base

bgamari:


glguy:
Looks like we should update the Travis-CI configuration to explicitly mention GHC 8.2.1 now as well as HEAD

hvr:
@glguy Fwiw, GHC HEAD is already in the travis config... it's just GHC 8.2.1 that's missing

PR #103, opened 2017-10-31
System.Posix.Signals: add haddock on signals

chris-martin:
The signal descriptions are from http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html


hvr:
@chris-martin That's awesome! Thanks for going the trouble to do this!

PR #107, opened 2018-02-26
Add comment why `closeFd` doesn't have to retry on EINTR

### LINKED ISSUE #88: closeFd doesn't check for EINTR
From `man 2 open`:

`EINTR  While blocked waiting to complete an open of a slow device (e.g., a FIFO; see fifo(7)), the call was interrupted by a signal handler; see signal(7).`

From `man 2 close`:

`EINTR  The close() call was interrupted by a signal; see signal(7).`

The situation in `unix`:

* For `open` [the code doesn't handle `EINTR`](https://github.com/haskell/unix/blob/7b20b4cbc35a2fea3f26c988f9b2a95640babaaa/System/Posix/IO/Common.hsc#L166)
* For `close` [the code doesn't handle `EINTR`](https://github.com/haskell/unix/blob/7b20b4cbc35a2fea3f26c988f9b2a95640babaaa/System/Posix/IO/Common.hsc#L194)

nh2:
Fixes #88

### DIFF
--- a/System/Posix/IO/Common.hsc
+++ b/System/Posix/IO/Common.hsc
@@ -221,6 +221,12 @@ foreign import capi unsafe "HsUnix.h open"
 
 closeFd :: Fd -> IO ()
 closeFd (Fd fd) = throwErrnoIfMinus1_ "closeFd" (c_close fd)
+-- Here we don't to retry on EINTR because according to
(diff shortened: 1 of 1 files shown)


hvr:
thanks!

PR #110, opened 2018-05-08
Bind `fdopendir` and `openat`

strake:


hvr:
Sorry haven't yet had time to review this. I'm a bit worried about two things here: a) routing everything through `openat(2)` and thus replace `open(2)` and b) these are POSIX.1-2008 entry points 

Are we sure all non-EOL'ed target platforms provide/support `openat(2)` and `fdopendir(3)`?

strake:
The package description specifies POSIX.1-2008.

Is somewhere a list of supported platforms of this package? I checked all supported Unix platforms of GHC and they all have both `openat` and `fdopendir` — indeed, both GNU libc on Linux and FreeBSD libc now call `openat` rather than `open` (not sure about OS X).

hvr:
I only mentioned POSIX.1-2008 because this means these could be recent enough that some less mainstream platforms we support may not have supported them yet. And if that's the case, we'd need to consider conditional compilation & autoconf tests.

Unfortunately I don't have a comprehensive list of supported platforms but we also support e.g. AIX, Solaris, OpenBSD and Android

strake:
PTAL

strake:
ping

strake:
ping

strake:
@hvr ping

System/Posix/Directory/Common.hsc
```diff
@@ -22,6 +22,9 @@
 
 module System.Posix.Directory.Common (
        DirStream(..), CDir, CDirent, DirStreamOffset(..),
+#ifdef HAVE_FDOPENDIR
+       openDirStreamFd,
```
hasufell: IMO, this should be in its own module `System.Posix.Directory.Fd`, no?

System/Posix/Directory/Common.hsc
```diff
@@ -36,12 +39,23 @@ module System.Posix.Directory.Common (
...
 import Foreign.C
+import Data.Functor ((<$>))
 
 newtype DirStream = DirStream (Ptr CDir)
 
 data {-# CTYPE "DIR" #-} CDir
 data {-# CTYPE "struct dirent" #-} CDirent
 
+#ifdef HAVE_FDOPENDIR
+-- | @openDirStreamFd fd@ calls @fdopendir@ to obtain a
```
hasufell: This should document that you must not close the file descriptor afterwards with a [link to the POSIX documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopendir.html).

System/Posix/Directory/Common.hsc
strake: Done

System/Posix/Directory/Common.hsc
strake: Done

hasufell:
@Rufflewind @hvr ping

hasufell:
@hvr ping

cartazio:
@chessai @hvr  @bodigrim This seems like a pretty decent patch, what are the support surface area concerns blocking it at the moment?

like, what OS's does the haskell Unix need to support, and whats missing to make this or something similar a good patch?

hasufell:
Guys, why is no one responding here? It's discouraging for contributors to wait 2 years for actionable input.

System/Posix/IO.hsc
```diff
@@ -31,7 +31,7 @@ module System.Posix.IO (
     -- ** Opening and closing files
     OpenMode(..),
     OpenFileFlags(..), defaultFileFlags,
-    openFd, createFile,
+    openFd, openFdAt, createFile, createFileAt,
```
Bodigrim: Should not `openFdAt` and `createFileAt` be guarded by `#ifdef HAVE_OPENAT`?

System/Posix/IO.hsc
```diff
@@ -72,6 +72,35 @@ import System.Posix.Error
...
+#ifdef HAVE_OPENAT
+-- |Open and optionally create this file.  See 'System.Posix.Files'
+-- for information on how to use the 'FileMode' type.
+openFd :: FilePath
+       -> OpenMode
+       -> OpenFileFlags
+       -> IO Fd
+openFd = openFdAt Nothing
+
+openFdAt :: Maybe Fd
```
Bodigrim: It would be nice to add a haddock comment for `openFdAt`.

System/Posix/IO.hsc
```diff
@@ -72,6 +72,35 @@ import System.Posix.Error
...
+     throwErrnoPathIfMinus1Retry "openFdAt" name $
+       openat_ fdMay str how flags
+
+-- |Create and open this file in WriteOnly mode.  A special case of
+-- 'openFd'.  See 'System.Posix.Files' for information on how to use
+-- the 'FileMode' type.
+createFile :: FilePath -> FileMode -> IO Fd
+createFile = createFileAt Nothing
+
+createFileAt :: Maybe Fd -> FilePath -> FileMode -> IO Fd
```
Bodigrim: It would be nice to add a haddock comment for `createFileAt`.

System/Posix/IO/ByteString.hsc
```diff
@@ -72,6 +72,35 @@ import System.Posix.IO.Common
...
+#ifdef HAVE_OPENAT
+-- |Open and optionally create this file.  See 'System.Posix.Files'
+-- for information on how to use the 'FileMode' type.
+openFd :: RawFilePath
+       -> OpenMode
+       -> OpenFileFlags
+       -> IO Fd
+openFd = openFdAt Nothing
+
+openFdAt :: Maybe Fd
```
Bodigrim: It would be nice to add a haddock comment for `openFdAt`.

System/Posix/IO/ByteString.hsc
```diff
@@ -72,6 +72,35 @@ import System.Posix.IO.Common
...
+     throwErrnoPathIfMinus1Retry "openFdAt" name $
+       openat_ fdMay str how flags
+
+-- |Create and open this file in WriteOnly mode.  A special case of
+-- 'openFd'.  See 'System.Posix.Files' for information on how to use
+-- the 'FileMode' type.
+createFile :: RawFilePath -> FileMode -> IO Fd
+createFile = createFileAt Nothing
+
+createFileAt :: Maybe Fd -> RawFilePath -> FileMode -> IO Fd
```
Bodigrim: It would be nice to add a haddock comment for `createFileAt`.

System/Posix/IO/ByteString.hsc
```diff
@@ -31,7 +31,7 @@ module System.Posix.IO.ByteString (
     -- ** Opening and closing files
     OpenMode(..),
     OpenFileFlags(..), defaultFileFlags,
-    openFd, createFile,
+    openFd, openFdAt, createFile, createFileAt,
```
Bodigrim: Should not `openFdAt` and `createFileAt` be guarded by `#ifdef HAVE_OPENAT`?

cartazio:
thank you for getting this rolling @Bodigrim

hasufell:
Please double check this code on mac!

I have a similar code: https://github.com/hasufell/hpath/blob/master/hpath-posix/src/System/Posix/RawFilePath/Directory/Traversals.hs#L225

And `fdopendir` usage here causes `d_name` to have the first char truncated. I'm not sure why. The use of `unsafeCoerce` seems safe (just converting to newtype).

cartazio:
emphasizing what @hasufell says,

at least on OSX, a number of newer unix calls really need to be bound via capi (an implicit header aware c wrapper) or an explicit c call with the correct system headers included

reproducing the example also mentioned above  that i cooked up and shared with julian:
```haskell
foreign import capi unsafe "dirent.h fdopendir"
    c_fdopendir :: Posix.Fd -> IO (Ptr CDir)
```
the header part of the capi syntax is crucial for this to compile and link correctly! (its actually impossible to bind the correct underlying system call symbol directly)

l29ah:
wtf just merge it already

hasufell:
> wtf just merge it already

The code is broken. See above.

l29ah:
The code works on linux and doesn't break anything. If it's bugged on some other platform, it could be addressed separately as a bugfix. Meanwhile, MacOS doesn't even strive to be POSIX these days it seems: https://www.quora.com/Is-Mac-OS-X-more-POSIX-compliant-than-Linux/answer/Madars-Oz

cartazio:
Julian and I figured out how to correctly bind some of this stuff on OS X. But incorrectly binding it will make programs flat out crash and silently corrupt data. 

A pr that does the capi indirection would be safe to merge. But this pr is flat out wrong. Someone prepare a pr that does the capi stuff and then we can make it work

cartazio:
Cc @Bodigrim and victor

Bodigrim:
@l29ah currently the only OS excluded to use `unix` package is Windows:
https://github.com/haskell/unix/blob/f7b956cad4b2eb745d0278bf5029222d1c8dfc01/unix.cabal#L67-L71
So even if MacOS is not POSIX compliant, we cannot just silently break it.

l29ah:
This pull request doesn't break anything.

l29ah:
Also the function definition can be put under a conditional if you don't want to create a false hope for Mac users.

cartazio:
Bodigrim: it looks like the binding is capi based now. @hasufell what was the min version of mac we found that supported this call?

Bodigrim:
@l29ah Providing conditional API is an option, but so far `unix` avoided it, sticking to a role of a common denominator for all non-Windows platforms. One could position `unix` as a wrapper for POSIX - and woe to systems which are not, but I'd say that such approach would greatly diminish its utility.

hs-viktor:
I've rebased this PR on master and pushed two fixup commits to address the `CApiFFI` issue for MacOS and missing `#if` guards for `openat` functions.

All CI targets now pass. It just remains to do a final closer review, but we should be able to merge this (perhaps squashed) soon barring unexpected issues.

hs-viktor (commented):
This is almost done, but I am concerned about the conditionally defined functions.  I don't see how a dependent application or library can safely make use of these functions, in the sense that it will at least reliably compile on all platforms.  Do they need to also run configure and use `HAVE_OPENAT`, ...?

There should likely be some way to discover whether these are available at runtime. So some sort of stub implementation needs to exist, and some way to test whether the function is expected to work, or whether the caller should take an alternative code path (or give up).

Perhaps just simple boolean predicates that are exported along with these?

Unless we've reached the point where the only supported Posix platforms all have `openat`, et. al. and we just fail to build `unix` otherwise.  So that these are then always available.

hs-viktor:
> There should likely be some way to discover whether these are available at runtime. So some sort of stub implementation needs to exist, and some way to test whether the function is expected to work, or whether the caller should take an alternative code path (or give up).
> 
> Perhaps just simple boolean predicates that are exported along with these?

One simple way to handle these is to export a `Maybe T` where `T` is the type of the optionally implemented function, and not export the function itself.  A caller then write:
```haskell
 case getOpenFdAt of
  Just f -> f ...
  Nothing -> ...
```
which, with appropriate inlining, will optimise out the test and take just the appropriate branch.

System/Posix/Directory/Common.hsc
```diff
@@ -32,12 +35,24 @@ module System.Posix.Directory.Common (
...
 data {-# CTYPE "struct dirent" #-} CDirent
 
+#ifdef HAVE_FDOPENDIR
+-- | @unsafeOpenDirStreamFd fd@ calls @fdopendir@ to obtain a directory stream
+--   for @fd@. @fd@ must not be otherwise used after this; see
+--   <https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopendir.html POSIX>.
+unsafeOpenDirStreamFd :: Fd -> IO DirStream
+unsafeOpenDirStreamFd (Fd fd) = DirStream <$> throwErrnoIfNull "openDirStreamFd" (c_fdopendir fd)
+
+foreign import capi unsafe "dirent.h fdopendir"
```
hasufell, line 52: Let's add a comment saying that the reason using capi is an actual bug.

vdukhovni: Agreed, we don't want this accidentally reverted back.

hs-viktor:
The openat function was added in:

* Linux 2.6.16 with glibc 2.4  (Mar 2006)
* FreeBSD 8.0  (Nov 2009)
* OpenBSD 5.0 (Nov 2011)
* MacOS 10.10 (Oct 2014)
* NetBSD 7.0 (Aug 2015)

Has it been established long enough that we can just require it unconditionally?

hs-viktor:
The timeline for `fdopendir` seems to be:

* glibc 2.4
* FreeBSD 8.0
* OpenBSD 5.0
* MacOS 10.10
* NetBSD 6.0

Pretty much the same, but slightly older with NetBSD.  So if we can require `openat` outright, we could do the same with `fdopendir` and dispense with the CPP conditions.

Bodigrim:
> Has it been established long enough that we can just require it unconditionally?

Yes, I believe 5+ years are enough. It is better to enable it unconditionally now than have a breaking switch from `Maybe T` to `T` in future, or pay for `Maybe` infinitely long.

cartazio:
Sounds good to me

hs-viktor:
The latest commit drops the `#if` guards. Builds on all the platforms on which run CI. I also added missing haddocks. Please review.

Separately from this MR, I really dislike the way that the file open flags are handled. We only support a fixed subset, hard-code as a list of booleans in a structure.  This is IMHO awful. It screams out for pattern synonyms and a single numeric flag field, that users can set to values that are not yet known to the library, if they know what they're doing.

I should open a separate issue...

Bodigrim (approved):

System/Posix/Directory/Common.hsc
```diff
@@ -32,12 +35,24 @@ module System.Posix.Directory.Common (
...
 
 newtype DirStream = DirStream (Ptr CDir)
 
 data {-# CTYPE "DIR" #-} CDir
 data {-# CTYPE "struct dirent" #-} CDirent
 
+#ifdef HAVE_FDOPENDIR
+-- | @unsafeOpenDirStreamFd fd@ calls @fdopendir@ to obtain a directory stream
+--   for @fd@. @fd@ must not be otherwise used after this; see
+--   <https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopendir.html POSIX>.
```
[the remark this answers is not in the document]
hs-viktor: @Bodigrim I see you've approved the MR, but perhaps you did not notice the above issue/question amidst the flurry of activity to resolve the easier obstacles.  I do think we still need to do something here...

hs-viktor (changes requested):
Disposition of file descriptor passed to `fdopendir` when the call fails remains as yet unaddressed. Should it be automatically closed (I think so), but otherwise callers need to know that they have to catch the exception and do something with the descriptor.

Bodigrim:
I'm in favor of closing it automatically.

vdukhovni:
> I'm in favor of closing it automatically.

OK, I'll arrange to close it with `c_close`, but it is perhaps worth noting that if any of the descriptors being passed into the library are being used in some other thread with `threadWaitRead` or `threadWaitWrite`, then they have to be closed with `closeFdWith`, rather than just `c_close`, so that the IO Manager can do appropriate cleanup.

Nothing in `unix` uses the IO Manager interfaces, but we perhaps need to make it clear to users that they need to be careful with passing such descriptors into `unix` calls that might close the descriptor.  Not likely to be an issue with a directory handle, so I think it is fairly safe to not worry too much with `fdopendir`, but something that may warrant a closer look elsewhere...

hs-viktor:
Final cut:
* `unsafeOpenDirStreamFd` input descriptor never left unpainted
* squashed
* changelog updated

Looking at the surrounding code makes me wish for a time-machine to address other issues that will require some care, but not in this PR. :-(

Bodigrim (approved):
Looks good!

PR #115, opened 2018-06-26
allow building with time 1.9.x

peti:
New versions of time `time` compile fine with this code.


peti:
According to the changelog, all changes made in 1.9.x were additions to the API, so I think it's fair to assume that `unix` won't break.

hvr:
IIRC the changelog is incomplete for time-1.9, but I don't recall what changed (except that it was indeed something demanding a major version increment).

PR #119, opened 2018-08-01
Resource: Fix warning in case of no RLIM_SAVED_{CUR,MAX}

bgamari:
This previously failed to build with a warning on FreeBSD 11.2


hvr:
thanks!

PR #121, opened 2018-10-30
Allow time-1.9

bgamari:


dbaynard:
Hackage still shows `< 1.9`. Would a metadata revision be possible?

PR #122, opened 2018-11-11
Fix issue 60: confusing error messages in rename/createLink/createSymbolicLink

### LINKED ISSUE #60: Misleading error messages in `rename`

luispedro:
This fixes #60. It builds on #65 (I cherry-picked the commit), while addressing the issues raised there.

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -102,13 +102,18 @@ import System.Posix.Files.Common
 import System.Posix.Error
 import System.Posix.Internals
 
+#if !MIN_VERSION_base(4, 11, 0)
+import Data.Monoid ((<>))
+#endif
+
 import Data.Time.Clock.POSIX (POSIXTime)
 
 -- throwErrnoTwoPathsIfMinus1_
 --
 -- | For operations that require two paths (e.g., renaming a file)
+throwErrnoTwoPathsIfMinus1_ :: (Eq a, Num a) => String -> FilePath -> FilePath -> IO a -> IO ()
 throwErrnoTwoPathsIfMinus1_ loc path1 path2 =
-  throwErrnoIfMinus1_ (loc ++ " '" ++ path1 ++ "' to '" ++ path2 ++ "'")
+  throwErrnoIfMinus1_ (loc <> " '" <> path1 <> "' to '" <> path2 <> "'")
 
 -- -----------------------------------------------------------------------------
 -- chmod()
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -104,6 +104,12 @@ import System.Posix.Internals
 
 import Data.Time.Clock.POSIX (POSIXTime)
 
+-- throwErrnoTwoPathsIfMinus1_
+--
+-- | For operations that require two paths (e.g., renaming a file)
+throwErrnoTwoPathsIfMinus1_ loc path1 path2 =
+  throwErrnoIfMinus1_ (loc ++ " '" ++ path1 ++ "' to '" ++ path2 ++ "'")
+
 -- -----------------------------------------------------------------------------
 -- chmod()
 
@@ -228,7 +234,7 @@ createLink :: FilePath -> FilePath -> IO ()
 createLink name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoIfMinus1_ ("createLink (target: "++name1++") (linkpath: "++name2++")") (c_link s1 s2)
+  throwErrnoTwoPathsIfMinus1_ "createLink" name1 name2 (c_link s1 s2)
 
 -- | @removeLink path@ removes the link named @path@.
 --
@@ -252,7 +258,7 @@ createSymbolicLink :: FilePath -> FilePath -> IO ()
 createSymbolicLink name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoIfMinus1_ ("createSymbolicLink (target: "++name1++") (linkpath: "++name2++")") (c_symlink s1 s2)
+  throwErrnoTwoPathsIfMinus1_ "createSymbolicLink" name1 name2 (c_symlink s1 s2)
 
 foreign import ccall unsafe "symlink"
   c_symlink :: CString -> CString -> IO CInt
@@ -290,7 +296,7 @@ rename :: FilePath -> FilePath -> IO ()
 rename name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoIfMinus1_ ("rename "++name1++" to "++name2) (c_rename s1 s2)
+  throwErrnoTwoPathsIfMinus1_  "rename" name1 name2 (c_rename s1 s2)
 
 foreign import ccall unsafe "rename"
    c_rename :: CString -> CString -> IO CInt
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -228,7 +228,7 @@ createLink :: FilePath -> FilePath -> IO ()
 createLink name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoIfMinus1_ ("createLink "++name1++" to "++name2) (c_link s1 s2)
+  throwErrnoIfMinus1_ ("createLink (target: "++name1++") (linkpath: "++name2++")") (c_link s1 s2)
 
 -- | @removeLink path@ removes the link named @path@.
 --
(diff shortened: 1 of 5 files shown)


System/Posix/Files.hsc
```diff
@@ -229,7 +229,7 @@ createLink :: FilePath -> FilePath -> IO ()
 createLink name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createLink" name1 (c_link s1 s2)
+  throwErrnoIfMinus1_ ("createLink (target: "++name1++") (linkpath: "++name2++")") (c_symlink s1 s2)
```
hs-viktor, line 232: This should not change the function from creating a hard link to creating a symbolic link:
```suggestion
  throwErrnoIfMinus1_ ("createLink (target: "++name1++") (linkpath: "++name2++")") (c_link s1 s2)
```

System/Posix/Files.hsc
```diff
@@ -253,7 +253,7 @@ createSymbolicLink :: FilePath -> FilePath -> IO ()
 createSymbolicLink file1 file2 =
   withFilePath file1 $ \s1 ->
   withFilePath file2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createSymbolicLink" file2 (c_symlink s1 s2)
+  throwErrnoIfMinus1_ ("createSymbolicLink (target: "++file1++") (linkpath: "++file2++")") (c_symlink s1 s2)
```
hs-viktor, line 256: Should we be consistent and also use `name1` and `name2` here, just in the other functions?

hs-viktor (changes requested):
Please also rebase on master, so we run this through the most recent CI incarnation.

luispedro:
Previous issues have been address and the branch has been rebased against `main`. CI looks good too.

System/Posix/Files.hsc
vdukhovni: This does not appear to be fixed. Why is it *still* calling `c_symlink` here?

vdukhovni (changes requested):

vdukhovni (approved):

luispedro:
I'm so sorry, I really thought I had addressed that. Fixed now (and I will add a test)

luispedro:
Added a test, which would have caught the previously mistake of creating symlinks when normal links were desired

tests/FileStatus.hs
```diff
@@ -63,10 +65,35 @@ testSymlink fs ds = do
...
+      actualF  = snd (statusElements ls)
+
+  when (actualF /= expected) $
+    fail "unexpected file status bits for hard link to regular file"
+
+  when (linkCount fs' /= 2) $
+    fail "newly created hard link was expected to contain have a link count of 2"
+
+  when (statusElements fs /= statusElements fs') $
+    fail "status for a file does not match when it's accessed via a link"
```
Bodigrim, line 90: Since you expanded this test suite, I think it's high time to improve its UX. Let's use [`assertEqual`](http://hackage.haskell.org/package/tasty-hunit-0.10.0.3/docs/Test-Tasty-HUnit.html#v:assertEqual) or [`@?=`](http://hackage.haskell.org/package/tasty-hunit-0.10.0.3/docs/Test-Tasty-HUnit.html#v:-64--61--63-) to report what exactly mismatches. You can find an example in [`ResourceLimit`](https://github.com/haskell/unix/blob/cdff32fd57cc5a1e6d7f1c08a9e4ea390284d9c8/tests/ResourceLimit.hs) test.

System/Posix/Files.hsc
```diff
@@ -290,7 +290,7 @@ rename :: FilePath -> FilePath -> IO ()
 rename name1 name2 =
   withFilePath name1 $ \s1 ->
   withFilePath name2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "rename" name1 (c_rename s1 s2)
+  throwErrnoIfMinus1_ ("rename "++name1++" to "++name2) (c_rename s1 s2)
```
Bodigrim, line 293: Filenames can include spaces, I usually find such messages more readable and unambiguous when paths are enclosed in single quotes. 

How about a shared helper function for `rename` / `createLink` / `createSymbolicLink`?
```haskell
throwErrnoTwoPathsIfMinus1_ loc path1 path2 = 
  throwErrnoIfMinus1_ (loc ++ "'" ++ path1 ++ "' to '" ++ path2 ++ "'")
```

tests/FileStatus.hs
luispedro: Done for the newer test. I think it's outside of the scope of this PR to improve the other tests, but it can serve as an example

System/Posix/Files.hsc
luispedro: I have no strong opinion on these issues. I'm happy to just do it as you prefer.

tests/FileStatus.hs
```diff
@@ -70,24 +71,16 @@ testLink = do
...
-
-  when (actualF /= expected) $
-    fail "unexpected file status bits for hard link to regular file"
-
-  when (linkCount fs' /= 2) $
-    fail "newly created hard link was expected to contain have a link count of 2"
-
-  when (statusElements fs /= statusElements fs') $
-    fail "status for a file does not match when it's accessed via a link"
+  linkCount fs' == 2 @? "Newly created hard link was expected to contain have a link count of 2"
```
Bodigrim, line 82: It's probably either "to contain" or "to have"?..

luispedro: good catch, yes. will fix in next push

Bodigrim (approved):

System/Posix/Files.hsc
```diff
@@ -104,6 +104,12 @@ import System.Posix.Internals
 
 import Data.Time.Clock.POSIX (POSIXTime)
 
+-- throwErrnoTwoPathsIfMinus1_
+--
+-- | For operations that require two paths (e.g., renaming a file)
+throwErrnoTwoPathsIfMinus1_ loc path1 path2 =
```
hs-viktor, line 110:
```suggestion
throwErrnoTwoPathsIfMinus1_ :: String -> FilePath -> FilePath -> IO ()
throwErrnoTwoPathsIfMinus1_ loc path1 path2 =
```

System/Posix/Files.hsc
```diff
@@ -104,6 +104,12 @@ import System.Posix.Internals
...
+  throwErrnoIfMinus1_ (loc ++ " '" ++ path1 ++ "' to '" ++ path2 ++ "'")
```
hs-viktor, line 111:
```suggestion
  throwErrnoIfMinus1_ (loc <> " '" <> path1 <> "' to '" <> path2 <> "'")
```
Just in case we ever move a `String` type that is not a `[Char]`, but something more like `UTF-8` text.  The more polymorphic `<>` should work even if `String` is no longer a list.  (Still holding out hope that could happen some day).

hs-viktor (changes requested):
Actually, I just noticed that similar changes also need to happen in `System/Posix/Files/ByteString.hsc`.  Please make these consistent.  There the filenames will be ByteStrings (`RawFilePath`), and throw errors via `System.Posix.ByteString.FilePath`.

hs-viktor:
> Actually, I just noticed that similar changes also need to happen in `System/Posix/Files/ByteString.hsc`. Please make these consistent. There the filenames will be ByteStrings (`RawFilePath`), and throw errors via `System.Posix.ByteString.FilePath`.

The new function for two paths should probably be added to that module and its non-bytestring twin.

System/Posix/Files.hsc
```diff
@@ -104,6 +104,13 @@ import System.Posix.Internals
 
 import Data.Time.Clock.POSIX (POSIXTime)
 
+-- throwErrnoTwoPathsIfMinus1_
+--
+-- | For operations that require two paths (e.g., renaming a file)
+throwErrnoTwoPathsIfMinus1_ :: (Eq a, Num a) => String -> FilePath -> FilePath -> IO a -> IO ()
+throwErrnoTwoPathsIfMinus1_ loc path1 path2 =
+  throwErrnoIfMinus1_ (loc <> " '" <> path1 <> "' to '" <> path2 <> "'")
```
hs-viktor, line 112: I should probably have mentioned that `(<>)` is not in `Prelude` prior to GHC 8.4, but we support 8.2, so you'll need:
```haskell
#if !MIN_VERSION_base(4,11,0)
import Data.Semigroup ( (<>) )
#endif

Comment (luispedro) [reply System/Posix/Files.hsc:112]: Yeah, I just ran into that :) I will push a new update in a few ...

**Comment by luispedro** [comment]
Updated the code to use `<>` & have a signature.

The `ByteString` module stuff might have to wait until the evening as it seems a bit more involved..

**Comment by luispedro** [comment]
`ByteString` module updated as well (including tests)

**Review by hs-viktor — APPROVED** [review]
Very nice. Thanks.

**Review by Bodigrim — APPROVED** [review]

File: System/Posix/Files.hsc [hunk 9]
```diff
@@ -241,18 +252,18 @@ removeLink name =
...
 -- Note: calls @symlink@.
 createSymbolicLink :: FilePath -> FilePath -> IO ()
-createSymbolicLink file1 file2 =
-  withFilePath file1 $ \s1 ->
-  withFilePath file2 $ \s2 ->
-  throwErrnoPathIfMinus1_ "createSymbolicLink" file2 (c_symlink s1 s2)
+createSymbolicLink name1 name2 =
+  withFilePath name1 $ \s1 ->
+  withFilePath name2 $ \s2 ->
+  throwErrnoTwoPathsIfMinus1_ "createSymbolicLink" name1 name2 (c_symlink s1 s2)
```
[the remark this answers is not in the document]
nh2: @vdukhovni Thanks for the example!

I agree, we definitely should populate `ioe_filename` as you say.

PR #123, opened 2018-11-19
Add missing blockSize and blockCount accessors to System.Posix.Files.

puvar:
These fields are specified by POSIX (see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html), but are currently missing.

This change depends on my change to System.Posix.Types: https://phabricator.haskell.org/D5354

### DIFF
--- a/System/Posix/Files.hsc
+++ b/System/Posix/Files.hsc
@@ -60,6 +60,7 @@ module System.Posix.Files (
     specialDeviceID, fileSize, accessTime, modificationTime,
     statusChangeTime,
     accessTimeHiRes, modificationTimeHiRes, statusChangeTimeHiRes,
+    blockSize, blockCount,
     isBlockDevice, isCharacterDevice, isNamedPipe, isRegularFile,
     isDirectory, isSymbolicLink, isSocket,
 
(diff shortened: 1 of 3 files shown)


hvr:
Well, this requires some careful consideration about backward compat with older `base` versions

Notice the shape of the `unix` compat matrix at https://matrix.hackage.haskell.org/#/package/unix

hvr:
Duplicate of #78

puvar:
If backwards compat is a concern, we can do either: (a) return CBlkSize and CBlkCnt directly, or (b) define BlockSize and BlockCount types in System.Posix.Files. When the new types appear in System.Posix.Types, we can switch to them; it will not break too many things (only for people who explicitly used value constructors in their apps).

As for the XSI concern in diff #78, I don't think we should check for existence of these fields in stat struct. They are present in every POSIX system I'm aware of. XSI is an extension, but it's an extension within POSIX standard, so any POSIX compliant system tends to implement it. So should any library that wraps around POSIX interface.

Bodigrim:
This is blocked until a) either we drop support of older GHCs, b) or agree our stance on conditional APIs, c) fix backward compatibility in other way. 

At the moment all builds are failing, even with newer GHCs.

hs-viktor:
Should be rebased, and then we can see whether any issues remain now that #175 is merged...

hs-viktor:
With #78 merged, I'm closing this PR.  The only question is whether perhaps the names suggested here `blockSize, blockCount` are improvements on `fileBlockSize, fileBlocks` in #78.  If so, we could still change that before a release is cut. My take is that the names in #78 look OK, but I hold no strong convictions on the matter.

PR #128, opened 2019-04-26
Use FilePath instead of String

apeyroux:
withDL and withDL_ use String instead of String. This would be to keep coherence with dlopen.


hvr:
thanks

Issue #126, opened 2019-01-23
Change return type of `exitImmediately`


erikd:
Currently `exitImmediately` has a type:
```
exitImmediately :: ExitCode -> IO ()
```
and the documentation explicitly states that it never returns.

*Since* it never returns, it makes more sense for the type to be:
```
exitImmediately :: ExitCode -> IO a
```
to match the type of `System.Exit.exit` and related functions.

In addition, it would be good to extend the docs to state that unlike the `System.Exit.exit` family of functions, `exitImmediately` does not work via the Haskell exception system and hence, `exitImmediately` cannot be caught with an exception handler.

hvr:
Makes sense; however generalizing the type-signature this way does represent a breaking change in the presence of class constraints on the return type; so this would be unix-2.8 material

PR #130, opened 2019-05-04
Change the return type of exitImmediately

erikd:
New type is: exitImmediately :: ExitCode -> IO a

Closes: https://github.com/haskell/unix/issues/126

### DIFF
--- a/System/Posix/Process/Common.hsc
+++ b/System/Posix/Process/Common.hsc
@@ -383,9 +383,14 @@ readWaitStatus wstatp = do
 
 -- | @'exitImmediately' status@ calls @_exit@ to terminate the process
 --   with the indicated exit @status@.
---   The operation never returns.
-exitImmediately :: ExitCode -> IO ()
-exitImmediately exitcode = c_exit (exitcode2Int exitcode)
+--   The operation never returns. Since it does not use the Haskell exception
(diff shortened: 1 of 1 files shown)


erikd:
Not totally happy with solution. Happy to get suggestions for improvements.

System/Posix/Process/Common.hsc
```diff
@@ -383,9 +383,14 @@ readWaitStatus wstatp = do
...
-exitImmediately :: ExitCode -> IO ()
-exitImmediately exitcode = c_exit (exitcode2Int exitcode)
+--   The operation never returns. Since it does not use the Haskell exception
+--   system and it cannot be caught.
+exitImmediately :: ExitCode -> IO a
+exitImmediately exitcode = do
+	_ <- c_exit (exitcode2Int exitcode)
+	-- The above will exit the program, but need the following to satisfy
+	-- the type signature.
+	exitFailure
```
ggreif: I think you can recursively call `exitImmediately status` here. This code should be discovered as dead, so everything of the right type will do.

erikd: That is a nice idea

erikd:
Happier with now after applying @ggreif 's suggestions.

PR #133, opened 2019-06-08
Changes for GHC 8.8

bgamari:
Bump bound on `base` and mark a fragile test as such.


RyanGlScott:
These changes should also go into the `2.7` branch, right? I think that's what GHC HEAD is currently targeting anyway.

bgamari:
Correct, but I first want to get them into `master` to avoid inconsistency.

bgamari:
Actually, now that I think of it ed2cf61 probably shouldn't be merged to `2.7`: the `fragile` testsuite modified is not present on `ghc-8.8.

PR #135, opened 2019-10-21
Improve baud rate handling.

TravisWhitaker:
Currently this package provides support for the baud rates stipulated by POSIX (0 - 38400), and conditionally supports the common extensions 57600 and 115200.

Here I add more commonly available extended baud rates, namely those available in the current `termbits.h` Linux header.

### DIFF
--- a/System/Posix/Terminal/Common.hsc
+++ b/System/Posix/Terminal/Common.hsc
@@ -353,6 +353,7 @@ withMinInput termios count = unsafePerformIO $ do
     pokeElemOff c_cc (#const VMIN) (fromIntegral count :: CCc)
 
 data BaudRate
+  -- These are the standard rates stipulated by POSIX:
   = B0
   | B50
   | B75
@@ -369,8 +370,22 @@ data BaudRate
   | B9600
   | B19200
   | B38400
+  -- These are non-standard rates that are often present on modern Unixes:
   | B57600
   | B115200
+  | B230400
+  | B460800
+  | B500000
+  | B576000
+  | B921600
+  | B1000000
+  | B1152000
+  | B1500000
+  | B2000000
+  | B2500000
+  | B3000000
+  | B3500000
+  | B4000000
 
 inputSpeed :: TerminalAttributes -> BaudRate
 inputSpeed termios = unsafePerformIO $ do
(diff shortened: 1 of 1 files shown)


cartazio:
1) would be nice to have some tests. 
2) might be good to document what hardware or os combos explicitly 
Support these rates?

TravisWhitaker:
1. The only thing we can really test here without serial hardware is `word2Baud . baud2Word === id`. If that's worth checking or if others have ideas for good tests I'd be happy to write them.

2. Recent Linux supports all of these baud rates. I know recent macOS supports at least 921600, but I'm not sure exactly what it provides. I had thought about using CPP to provide a `BaudRate` type that only has constructors for the rates supported on the current platform. It'd be nice to get a build-time error if a missing baud rate is used, but this didn't seem worth the chaos that I can imagine conditionally-present-constructors might cause.

cartazio:
just to clarify: by baud rate we mean serial port interface?

TravisWhitaker:
Hey @bgamari @hvr @RyanGlScott what are the next steps for this PR?

TravisWhitaker:
Pinging @bgamari @hvr @RyanGlScott how can this be moved forward?

hvr:
I'll merge this for now (thus this will be part of the next major version of `unix` which isn't imminent anyway; so there's still some time to figure out a better API before this ends up in a release). I'm generally not to happy about the current API (which this patch is effectively just adhering to and extending) which doesn't allow a good way to probe whether a baudrate is even supported, and effectively needs users to potentially deal with two different failure codepaths (i.e. ioctl fails, vs. CPP macro wasn't visible at compile-time which results in a `error`-value when converting the enum constructor).

TravisWhitaker:
@hvr How would you feel about something like this?

```haskell
newtype BaudRate = BaudRate CSpeed

#ifdef B0
pattern B0 :: BaudRate
pattern B0 = BaudRate (#const B0)
#endif
#ifdef B50
pattern B50 :: BaudRate
pattern B50 = BaudRate (#const B50)
#endif
-- and so on...
```

The only pain point I can think of is support for older GHCs that don't allow bidirectional pattern synonyms. I'm not sure what the legacy support contract is for boot libraries; so maybe this isn't allowed?

PR #141, opened 2019-12-01
Add some basic type class instances for `OpenMode` and `OpenFileFlags`.

### LINKED ISSUE #75: OpenMode should derive some classes (Eq)
It seems reasonable to derive Eq for `OpenMode`.  If there are explicitly no objections I will submit a pull request deriving `Eq, Ord, Show, Read`.

typedrat:
I've written a few different programs that provide orphan instances for these types, and I think that adding some basic derived instances should be uncontroversial.

Closes #75.


chessai:
Could you please add a changelog entry?

chessai:
Also, any reason to leave out `Read`?

typedrat:
Done.

chessai:
CI passed on the previous commit, most recent was to the changelog. These instances should be uncontroversial. Merging.

Thanks @typedrat!

PR #145, opened 2020-01-26
Implement BaudRate with PatternSynonyms.

TravisWhitaker:
This is a sketch of a new way of handling baud rates. The idea is to make `BaudRate` an opaque type, and provide pattern synonyms based on the availability of baud rate macros from the local termios header. There are two downsides to this approach:

- Older GHC versions don't support `-XPatternSynonyms`.
- The only way to write a total function of `BaudRate` is to use `baud2Word`.

I'm responsible for lots of Haskell code wrapping C APIs, and this approach seems to work well. The [gl package](https://hackage.haskell.org/package/gl) uses the same approach for handling constants from C headers.

### DIFF
--- a/System/Posix/Terminal/Common.hsc
+++ b/System/Posix/Terminal/Common.hsc
@@ -1,4 +1,5 @@
 {-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE Trustworthy #-}
 
 -----------------------------------------------------------------------------
@@ -48,7 +49,100 @@ module System.Posix.Terminal.Common (
   minInput,
   withMinInput,
 
-  BaudRate(..),
+  BaudRate,
+#ifdef B0
+  pattern B0,
+#endif
+#ifdef B50
+  pattern B50,
+#endif
+#ifdef B75
+  pattern B75,
+#endif
+#ifdef B110
+  pattern B110,
+#endif
+#ifdef B134
+  pattern B134,
+#endif
+#ifdef B150
+  pattern B150,
+#endif
+#ifdef B200
+  pattern B200,
+#endif
+#ifdef B300
+  pattern B300,
+#endif
+#ifdef B600
+  pattern B600,
+#endif
+#ifdef B1200
+  pattern B1200,
+#endif
+#ifdef B1800
+  pattern B1800,
+#endif
+#ifdef B2400
+  pattern B2400,
+#endif
+#ifdef B4800
+  pattern B4800,
+#endif
+#ifdef B9600
+  pattern B9600,
+#endif
+#ifdef B19200
+  pattern B19200,
+#endif
+#ifdef B38400
+  pattern B38400,
+#endif
+#ifdef B57600
+  pattern B57600,
+#endif
+#ifdef B115200
+  pattern B115200,
+#endif
+#ifdef B230400
+  pattern B230400,
+#endif
+#ifdef B460800
+  pattern B460800,
+#endif
+#ifdef B500000
+  pattern B500000,
+#endif
+#ifdef B576000
+  pattern B576000,
+#endif
+#ifdef B921600
+  pattern B921600,
+#endif
+#ifdef B1000000
+  pattern B1000000,
+#endif
+#ifdef B1152000
+  pattern B1152000,
+#endif
+#ifdef B1500000
+  pattern B1500000,
+#endif
+#ifdef B2000000
+  pattern B2000000,
+#endif
+#ifdef B2500000
+  pattern B2500000,
+#endif
+#ifdef B3000000
+  pattern B3000000,
+#endif
+#ifdef B3500000
+  pattern B3500000,
+#endif
+#ifdef B4000000
+  pattern B4000000,
+#endif
   inputSpeed,
   withInputSpeed,
   outputSpeed,
@@ -74,6 +168,7 @@ module System.Posix.Terminal.Common (
 
 import Data.Bits
 import Data.Char
+import Data.Coerce
 import Foreign.C.Error ( throwErrnoIfMinus1, throwErrnoIfMinus1_ )
 import Foreign.C.Types
 import Foreign.ForeignPtr ( ForeignPtr, withForeignPtr, mallocForeignPtrBytes )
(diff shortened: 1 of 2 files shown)


TravisWhitaker:
@hvr curious to hear your thoughts.

hvr:
While this is definitely an elegant way the big downside here is that this would require PatternSynonyms from the compiler, which cuts off everything before GHC 7.8 (supporting GHC 7.8 will require some extra work btw; the PatternSynonyms impl was improved in GHC 7.10 which renders the current PR incompatible with GHC 7.8) :-/

See https://hackage.haskell.org/package/unix to get an idea of unix' track-record of backward compat.

I'm not sure if this API refinement would justify breaking compatibility with those compilers; but if we do this, we can rethink other places as well to take advantage of newer GHCs -- then there's a lot more to break to make this count!

hasufell:
>  we can rethink other places as well to take advantage of newer GHCs -- then there's a lot more to break to make this count!

Which shouldn't block this PR though.

Bodigrim:
This is blocked, until we decide to abandon support for GHC < 8.0. @TravisWhitaker would you like to make a case why the proposed approach is better than status quo?

hs-viktor:
> This is blocked, until we decide to abandon support for GHC < 8.0. @TravisWhitaker would you like to make a case why the proposed approach is better than status quo?

I worked extensively with Kazu Yamamoto in the `dns` and `network` packages migrating various interfaces, that previously used algebraic sums to model system constants, to use `PatternSynonyms` instead.

Pattern synonyms are definitely the way to go when the underlying enumerated type is "open", in the sense that new member elements are added from time to time, or existing member elements are platform-specific.  The pattern synonyms are also more efficient, since they directly capture the underlying numeric values passed to system calls, ...

New releases of `dns` and `network` no longer support GHC 7.x.  From my perspective, pattern synonyms are an indispensable building block of system-programming with Haskell, and GHC 7.x is in a sense `unusable`, by forcing an inappropriate data model on various APIs, causing routine change to become breaking changes, ...

So I would strongly advocate for dropping support for GHC 7.x as soon as absolutely possible. The older releases are no longer maintained. I just filed MRs fixing a critical issue in the 8.8,8.10,9.0 and master MVar versions, but the same bug is present in GHC 8.6 and at least as far back as 7.8.  These GHC versions are abandonware, and I see no reason to continue to support them in updated major releases of packages.  Any users clinging to GHC 7.x can use corresponding older releases of packages they need.

System/Posix/Terminal/Common.hsc
```diff
@@ -48,7 +49,100 @@ module System.Posix.Terminal.Common (
   minInput,
   withMinInput,
 
-  BaudRate(..),
+  BaudRate,
```
hs-viktor, line 52: I'd like to see this export list maintained in just one module.  Perhaps `System.Postfix.Terminal.Exports` with `Common` and `ByteString` both re-exporting `Exports` or something along these lines...

System/Posix/Terminal/Common.hsc
```diff
@@ -352,40 +447,132 @@ withMinInput termios count = unsafePerformIO $ do
...
-  | B2500000
-  | B3000000
-  | B3500000
-  | B4000000
+newtype BaudRate = BaudRate CSpeed
+
+#ifdef B0
+pattern B0 :: BaudRate
+pattern B0 = BaudRate (#const B0)
+#endif
```
hs-viktor, line 455: Here and elsewhere I'd like to see the baud rates that are not supported on the current platform mapped to a known unused value, likely `-1` will work, ... so that the `well-known` values are defined on all platforms, even if not actually mapped to usable baud rate.

The sentinel value should also have an explicit name (`UnsupportedBaudRate` or similar), which allows users to check whether the baud-rate they are looking for is supported or not.

hs-viktor: A sentinel value works much better than bottom

System/Posix/Terminal/Common.hsc
```diff
@@ -581,119 +768,61 @@ cc2Word Stop      = (#const VSTOP)
...
-#else
-baud2Word B4000000 = error "B4000000 not available on this system"
-#endif
+baud2Word = coerce
 
 -- And convert a word back to a baud rate
 -- We really need some cpp macros here.
 
 word2Baud :: CSpeed -> BaudRate
 word2Baud x = case x of
```
hs-viktor, line 777: This function should accept any number, whether known at the time the package was released, or not, and should not throw errors by default.  If the value is not actually supported by the terminal driver, the user will find out when the relevant ioctl fails.

hs-viktor: We should however also provide sensible `Show` and `Read` instances for these, examples can be found in the `network` package.

hs-viktor (changes requested):

System/Posix/Terminal/Common.hsc
TravisWhitaker: Seems fine to me, but what's the rationale?

System/Posix/Terminal/Common.hsc
TravisWhitaker: What do you think is defined to be bottom here? On a given platform, the undefined baud rate macros' patterns are also undefined.

System/Posix/Terminal/Common.hsc
TravisWhitaker: I agree.

System/Posix/Terminal/Common.hsc
vdukhovni: My point is that none of the values should be "error", they should instead be a non-error sentinel, that one can compare a given value with to determine whether it is supported or not.  The unsupported value can probably be (-1).

TravisWhitaker: With this PR, none of the values are "error."

vdukhovni: But the patterns should be defined unconditionally, even when the OS does not have a define for the name, in which case the name should map to a sentinel value that indicates an invalid Baud rate.

TravisWhitaker: Why is that better than the behavior one gets with C, where using an undefined baud macro would yield a compile time error instead?

vdukhovni: > Why is that better than the behavior one gets with C, where using an undefined baud macro would yield a compile time error instead?

Because, in an actual C application, the call site would have an `#Ifdef` the particular macro, and the code would compile on all platforms!  Here, we're providing a Haskell module that abstracts the C interface, but the caller is not expected to have feature-test `CPP` macros in their code, that's our job. And yet the caller's code must compile on all platforms.

Therefore, we must present the same set of pattern synonyms on all platforms, and allow run-time discovery of which ones are supported by the platform.

System/Posix/Terminal/Common.hsc
hs-viktor: With #175 merged, this PR can resume progress.  But we do need to sort out my prior comments. Are you ready/willing to get this back on track?  If not, I could perhaps take it over and get it across the line?

hasufell:
Is this missing the `COMPLETE` pragma?

hs-viktor:
Closing, replaced by #216

PR #151, opened 2020-04-05
Create DirStream using an Fd, using fdopendir

tsuraan:
Not an interesting change, but I have a system that's passing Fds around through AF_UNIX connections, and being able to pass a directory Fd through that would be much simpler for me than passing around the directory contents. I'm hoping this isn't too controversial :)

A quick example of how this can be used would be:
```
Prelude System.Posix> :m +System.Posix.IO
Prelude System.Posix.IO System.Posix> :m +System.Posix.Directory
Prelude System.Posix.IO System.Posix.Directory System.Posix> fd <- openFd "/" ReadOnly defaultFileFlags
Prelude System.Posix.IO System.Posix.Directory System.Posix> dir <- fdOpenDirStream fd
Prelude System.Posix.IO System.Posix.Directory System.Posix> readDirStream dir
"."
```

### DIFF
--- a/System/Posix/Directory.hsc
+++ b/System/Posix/Directory.hsc
@@ -34,6 +34,7 @@ module System.Posix.Directory (
    -- * Reading directories
    DirStream,
    openDirStream,
+   fdOpenDirStream,
    readDirStream,
    rewindDirStream,
    closeDirStream,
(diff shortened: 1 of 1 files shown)


tsuraan:
Hm, just noticed a PR from 2018 doing this and more. I'm not sure if it's merged in yet or not, but maybe this is redundant?

PR #167, opened 2020-12-20
Assemble a proper test suite

Bodigrim:
...which can be run by usual means instead of fiddling with GHC make.


Bodigrim:
* MacOS builds fail with a curious mismatch between shells, returned by `getUserEntryForName` and `getUserEntryForID`: `/bin/sh` vs. `/bin/bash`. We do not do anything fancy with the output `CPasswd`, so it is not out fault.
* FreeBSD build errors look more interesting. @hs-viktor any ideas?

hs-viktor:
> * MacOS builds fail with a curious mismatch between shells, returned by `getUserEntryForName` and `getUserEntryForID`: `/bin/sh` vs. `/bin/bash`. We do not do anything fancy with the output `CPasswd`, so it is not out fault.
> * FreeBSD build errors look more interesting. @hs-viktor any ideas?

Yes, the test suite is buggy.

* T8108 is a really bad idea to run in the same process as other tests. It causes a CPU limit error, because the background loops no longer terminate after the thread delay.  This then breaks the subsequent test when CPU limits are exhausted.  Also this test is extremely wasteful reading the /etc/passwd and /etc/group files in a very tight loop. Avoid if possible, or run briefly in a separate process.
* Separately the FdReadBuf001.hs program is liable to deallocate the write buffer before the writer is done (and does on FreeBSD).

I have fixes for both.

hs-viktor:
> * MacOS builds fail with a curious mismatch between shells, returned by `getUserEntryForName` and `getUserEntryForID`: `/bin/sh` vs. `/bin/bash`. We do not do anything fancy with the output `CPasswd`, so it is not out fault.

The MacOS issue could be an artefact of the sandboxing environment the CI runs in.  It could well have multiple entries for pid 0, the first of which could have "/bin/bash" for a shell.  The test is not portable.  There is no requirement for the results to match.  On my MacOS Catalina system they do:
```
Prelude System.Posix.User> getUserEntryForName "root"
UserEntry {userName = "root", userPassword = "*", userID = 0, userGroupID = 0, userGecos = "System Administrator", homeDirectory = "/var/root", userShell = "/bin/sh"}
Prelude System.Posix.User> getUserEntryForID 0
UserEntry {userName = "root", userPassword = "*", userID = 0, userGroupID = 0, userGecos = "System Administrator", homeDirectory = "/var/root", userShell = "/bin/sh"}
```
but other systems are free to do as they please...

Bodigrim:
> The MacOS issue could be an artefact of the sandboxing environment the CI runs in. 

Yeah, I suspect it is caused by 
https://github.com/haskell/unix/blob/8378b72671137ad41179e3794043fb77333f88f6/.github/workflows/ci.yml#L9-L11
I removed comparison of shells.

> I have fixes for both.

Yay, thanks. Cherry-picked into this PR.

vdukhovni:
After "sleeping on it", I am somewhat concerned that the breakage from T8108 continuing to run in the same process overlapping with subsequent tests may be uncovering an actual thread-safety issue?  (Or perhaps a FreeBSD bug?)
This may be worth investigating a bit further.

While the "real" T8108 in the GHC testsuite is never tested this way, perhaps there's something here that's worth a look...

vdukhovni:
> We can keep T8108 in a separate executable, so that it does not interfere with other tests. I do not have a FreeBSD machine to investigate.

I have a FreeBSD machine, so I'll take a closer look.  My point was that perhaps the fact that T8108 continuing to run along with the other tests and thereby breaking them is possibly worth a closer look. Maybe there's some residual issue there that's only uncovered by the more complex interactions that this exposes.

So yes, for now, keep going with T8108 left out, or separate, but we should try to get to the bottom of what's actually happening here...

hs-viktor:
OK, I think I've found the problem with T8108 and the "user001" test, and it is in fact exposing a bug in the `unix` library.
The issue is with:
```haskell
lock :: MVar ()  
lock = unsafePerformIO $ newMVar ()

getAllGroupEntries :: IO [GroupEntry]
#ifdef HAVE_GETGRENT
getAllGroupEntries =     
    withMVar lock $ \_ -> bracket_ c_setgrent c_endgrent $ worker []
    where worker accum =
              do resetErrno
                 ppw <- throwErrnoIfNullAndError "getAllGroupEntries" $
                        c_getgrent
                 if ppw == nullPtr
                     then return (reverse accum)
                     else do thisentry <- unpackGroupEntry ppw
                             worker (thisentry : accum)
```
This is using a stateful (walks the group file incrementally) `getgrent(3)`, but attempts to work around that with an `MVar` lock, which works if the other threads are also using the same locked function, but fails if instead they're using `getgrgid_r(3)`, ... which is thread-safe only against other users of the keyed `_r` functions, it is NOT thread safe against concurrent users of `getgrent(3)` in *the same* thread (taking turns).  The file handle used by `getgrent(3)` is in thread-local-storage, but GHC does many-to-one threading, alternating haskell threads within the same OS thread.  So the `T8108` threads mess up the file walks for `getAllGroupEntries`. 

Just leaving out `getAllGroupEntries` makes the `user001` test pass, but that's not the solution.  We should be using `getgrent_r(3)` at least on FreeBSD, and wherever else it is available.

The solution is perhaps to run `getAllGroupEntires` in a separate *bound* (`forkOS`) thread, and return the results from there.  That'll work (on FreeBSD) in the threaded runtime.  Not sure what we should do in the single-threaded runtime, we need to avoid interleaving calls to `getgrent(3)` and `getgrgid_r(3)` in *the same* thread, and `getgrent_r(3)` does not help, it only solves the buffer reuse problem, not the file positioning problem.

I don't know whether or how Linux avoids the issue, perhaps `getgrnam_r(3)` et. al. never read the same file as `getgrent(3)` and ignore the efficiency gains of the file already being open.

Anyway, it looks like there's a real thread-safety issue here...

hs-viktor:
Also, while I was not a fan of the proposal to make most of the library interfaces "safe" with compile-time opt-out, the various `getpw*` and `getgr*`functions are generically expensive even when reading local `/etc/passwd` and `/etc/group` files, and especially when the nsswitch indirects to LDAP and the like.  So in `System.Posix.User`, we really should probably use "safe" for the functions that do "database" lookups (not just getuid(2), geteuid(2), ...).

Bodigrim:
I suggest we separate concerns here. I'd restore T8108 as a separate test executable, mimicking the way it used to be tested, so that we can go ahead with this PR. Fixing `getgrent` and "safe" `System.Posix.User` could be done later.

hs-viktor:
> I suggest we separate concerns here. I'd restore T8108 as a separate test executable, mimicking the way it used to be tested, so that we can go ahead with this PR. Fixing `getgrent` and "safe" `System.Posix.User` could be done later.

OK.  Perhaps also separate "resourceLimit" into a separate test executable, or put it *last*.  It imposes a hard CPU time limit that is not necessarily satisfied when running the tests, and I don't think that's the intention. I think it is just testing the API.

I also (intermittently, so far just once), saw a failure of `posix006`, which expects `sleep 1` to always yield an epochTime increment of 1 or 2, but I don't think that's reliable...

Also when building the unix-tests with "-threaded", and running with "+RTS -N", I'm now seeing failures in "posix002" (which spawns `printenv` with a given environment, haven't yet tracked that down).

So bottom line, yes let's get the test-suite integrated, and then see how many bugs we can squash...

hs-viktor:
By the way, I have a work-in-progress that still needs a bit of polish that fixes the `System.Posix.User` issues, I'll file a PR for this when it looks ready.

tests/Test.hs
```diff
@@ -0,0 +1,255 @@
...
+  , fileStatus
+  , fileStatusByteString
+  , getEnvironment01
+  , getEnvironment02
+  , getGroupEntry
+  , getUserEntry
+  , processGroup001
+  , processGroup002
+  , queryFdOption01
+  , resourceLimit
```
hs-viktor, line 37: Best to move `resourceLImit` to the bottom of the list, so it does not prematurely interrupt the execution of subsequent tests with a CPU limit exceeded error.

hs-viktor (commented):
Looks reasonable. I'll tackle the `get(pw|gr)*` issues separately, after this is merged.
Another option for `resourceLimit` instead of moving it to the end of the list, is to also run it in a separate test executable.

hs-viktor (approved):
LGTM. Thanks.

emilypi (approved):
LGTM, thanks @Bodigrim

hs-viktor:
@hvr Would you like us to wait to merge this until you've also had a chance to review?

chessai (approved):

chessai:
I read through this the other day and forgot to approve. This is a great change

emilypi (approved):

emilypi:
@Bodigrim would you like to do the honors of merging?

Bodigrim:
Thanks everyone!

PR #169, opened 2020-12-22
Thread safety of passwd/group lookups

vdukhovni:
- Simplify target platform assumptions: either getpwnam_r(3)
  and getpwuid_r(3) exist, or we don't support password lookups.
  The legacy non-thread-safe variants should not be needed on
  any supported platforms.

- Change the passwd/group FFI glue functions from "unsafe" to
  "safe".  These are not low-latency APIs.

- Then, sadly add some complexity to the implementation, because
  on FreeBSD systems the get(pw|gr)(nam|[ug]id)_r(3)
  "thread-safe" functions are only safe in 1-to-1 thread models.
  They share the thread-specific open file handle with calls
  from get(gr|pw)ent(3), breaking Haskell's sharing of OS threads
  for multiple "green" threads.

  The work-around is to use bound threads for the "*ent(3)" functions
  on FreeBSD systems when in the threaded RTS, and otherwise use locks
  to serialise all passwd/group file access.

  Linux, MacOS, NetBSD and OpenBSD systems don't appear to require the same
  work-around, as they don't share the FreeBSD passwd "compat" and getgrent(3),
  [bugs](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=252094)

### DIFF
--- a/System/Posix/User.hsc
+++ b/System/Posix/User.hsc
@@ -298,19 +298,15 @@ data UserEntry =
  } deriving (Show, Read, Eq)
 
 --
--- getpwuid and getpwnam leave results in a static object. Subsequent
--- calls modify the same object, which isn't threadsafe. We attempt to
--- mitigate this issue, on platforms that don't provide the safe _r versions
---
--- Also, getpwent/setpwent require a global lock since they maintain
+-- getpwent/setpwent require a global lock since they maintain
 -- an internal file position pointer.
-#if !defined(HAVE_GETPWNAM_R) || !defined(HAVE_GETPWUID_R) || defined(HAVE_GETPWENT) || defined(HAVE_GETGRENT)
+#if defined(HAVE_GETPWENT) || defined(HAVE_GETGRENT)
 lock :: MVar ()
 lock = unsafePerformIO $ newMVar ()
 {-# NOINLINE lock #-}
 #endif
 
--- | @getUserEntryForID gid@ calls @getpwuid_r@ to obtain
+-- | @getUserEntryForID uid@ calls @getpwuid_r@ to obtain
 --   the @UserEntry@ information associated with @UserID@
 --   @uid@. This operation may fail with 'isDoesNotExistError'
 --   if no such user exists.
@@ -324,14 +320,6 @@ getUserEntryForID uid =
 foreign import capi unsafe "HsUnix.h getpwuid_r"
   c_getpwuid_r :: CUid -> Ptr CPasswd ->
                         CString -> CSize -> Ptr (Ptr CPasswd) -> IO CInt
-#elif HAVE_GETPWUID
-getUserEntryForID uid = do
-  withMVar lock $ \_ -> do
-    ppw <- throwErrnoIfNull "getUserEntryForID" $ c_getpwuid uid
-    unpackUserEntry ppw
-
-foreign import ccall unsafe "getpwuid"
-  c_getpwuid :: CUid -> IO (Ptr CPasswd)
 #else
 getUserEntryForID = error "System.Posix.User.getUserEntryForID: not supported"
 #endif
@@ -351,15 +339,6 @@ getUserEntryForName name =
 foreign import capi unsafe "HsUnix.h getpwnam_r"
   c_getpwnam_r :: CString -> Ptr CPasswd
                -> CString -> CSize -> Ptr (Ptr CPasswd) -> IO CInt
-#elif HAVE_GETPWNAM
-getUserEntryForName name = do
-  withCAString name $ \ pstr -> do
-    withMVar lock $ \_ -> do
-      ppw <- throwErrnoIfNull "getUserEntryForName" $ c_getpwnam pstr
-      unpackUserEntry ppw
-
-foreign import ccall unsafe "getpwnam"
-  c_getpwnam :: CString -> IO (Ptr CPasswd)
 #else
 getUserEntryForName = error "System.Posix.User.getUserEntryForName: not supported"
 #endif
--- a/System/Posix/User.hsc
+++ b/System/Posix/User.hsc
@@ -54,10 +54,11 @@ import Foreign.Ptr
 import Foreign.Marshal
 import Foreign.Storable
 
-#if !defined(HAVE_GETPWNAM_R) || !defined(HAVE_GETPWUID_R) || defined(HAVE_GETPWENT) || defined(HAVE_GETGRENT)
-import Control.Concurrent.MVar  ( MVar, newMVar, withMVar )
+#if defined(HAVE_GETPWENT) || defined(HAVE_GETGRENT)
+#if defined(freebsd_HOST_OS)
+import Control.Concurrent (runInBoundThread, rtsSupportsBoundThreads)
 #endif
-#ifdef HAVE_GETPWENT
+import Control.Concurrent.MVar ( MVar, newMVar, withMVar )
 import Control.Exception
 #endif
 import Control.Monad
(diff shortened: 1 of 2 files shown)


hs-viktor:
@zoulasc Sorry to bug you about Haskell issues, but do you have any thoughts on this from a BSD-perspective? Perhaps Haskell is not the only many-to-1 threading ecosystem that is potentially affected by the unexpected interaction between `getpwnam_r(3)` (which one expects to not have side-effects on other activity) and `getpwent_r(3)`. It is clear that locks are needed to preclude two instances of `getpwent_r(3)` sharing the same thread, but not at all obvious that one needs to either also lock out `getpwnam_r(3)` or force `getpwent_r(3)` into a dedicated thread...

FWIW, I think that the `setpassent(3)` and `setgroupent(3)` optimizations are no longer a good idea, and that it might make sense to ignore them silently and always re-open the files in `get(pw|gr)(nam|[ug]id)_r(3)`, avoiding questions about stale files and unexpected interactions with `getpwent(3)`.  Of course any such change will take years to get adopted, so the Haskell libraries still need work-arounds, but perhaps *some day* a decade or more from now the issue will be moot...

zoulasc:
Unfortunately I think it is all bad news for get*ent(). Linux says they are not MT-Safe. The fget*ent() is documented to be MT-Safe, but that means only for the file backend? At work we are using an LDAP backend with SSSD and we have disabled enumeration because as suggested by the SSSD folks it can lead to performance issues (and that's very true). I think that with some amount of work each back-end can be made thread-safe and enabled for enumeration, but this is not the case today. The remedy seems to be: hand-hurts, cut-hand. Unless you cannot avoid using them I would stay clear. I have not looked at the BSD code yet.

vdukhovni:
> Unfortunately I think it is all bad news for get_ent(). Linux says they are not MT-Safe. [...] I have not looked at the BSD code yet.

Yes, indeed the Linux functions are not MT-safe, so we (Haskell/unix) have a mutex around these, which is sufficient on Linux.  The problem is that on BSD systems, we also either need a mutex around `getpwnam_r(3)` et. al. or need to arrange for `getent()` to run in a separate bound thread.  The reason is the surprising interaction between `getent` and `getnam()` on BSD, they can't take turns sharing the same thread, because the thread-local file-descriptor is used by both.

So that's what my question is about, should BSD perhaps consider not using any already open file in the `get(pw|gr)nam()` or `get(pw|gr)[ug]id` functions. I think that would be the right course of action, but of course there's no immediate benefit to downstream libraries, because the deployed base will change very slowly.

Nico suggests just deprecating enumeration, and I see the point, but am reluctant to pull the plug...

I could return just the entries for "root" and the user making the call, but sometimes half an answer is worse than none.  But also returning an empty result or throwing an error have drawbacks. All the choices are bad...

zoulasc:
I am with Nico here (stop supporting enumeration). I don't like it either, but I think it is going to be hard to fix everything. For example it seems that NetBSD is broken because it is using global state here: http://bxr.su/NetBSD/lib/libc/gen/getpwent.c#605. There could be other issues. I will take a look at all of it when I feel better.

hs-viktor:
> I am with Nico here (stop supporting enumeration). I don't like it either, but I think it is going to be hard to fix everything. For example it seems that NetBSD is broken because it is using global state here: http://bxr.su/NetBSD/lib/libc/gen/getpwent.c#605. There could be other issues. I will take a look at all of it when I feel better.

Thanks for the link. Actually, there's some good news for NetBSD here, it looks like the interaction with `stay open` and `getpwent` only happens in `get-wham()`, but, unlike FreeBSD, the NetBSD `getpwnam_r` does not share that state and is fully thread-safe. The manpage looks the same as on FreeBSD but the behaviour is less problematic.

On NetBSD (at least the version you pointed me at), I don't have to jump through hoops isolating `getpwnam_r` from the "net" functions, and probably don't need to use bound threads.

Of course we (Haskell/unix) may still decide to just drop enumeration support. I'll wait for the rest of the gang to pitch in...

hs-viktor:
I'm removing NetBSD from the list of OS platforms that need a FreeBSD-style work-around, the implementation is essentially unchanged since 2004: http://cvsweb.netbsd.org/bsdweb.cgi/src/lib/libc/gen/getpwent.c?only_with_tag=MAIN#rev1.57 and `getpnam_r` and friends don't have unexpected interactions with `get-went()`.

hs-viktor:
Here's a test Perl one-liner:
```Perl
$\="\n"; setpwent(); for ($i = 0; $i < 3; ++$i) { getpwuid(0); print join(":", getpwent()); } endpwent()
```
On FreeBSD this prints (ad nauseam, if the loop bound is not capped at 3):
```
toor:*:0:0:0::Bourne-again Superuser:/root::0
toor:*:0:0:0::Bourne-again Superuser:/root::0
toor:*:0:0:0::Bourne-again Superuser:/root::0
...
```
On NetBSD, I get:
root:*:0:0:0::Charlie &:/root:/bin/sh:0
toor:*:0:0:0::Bourne-again Superuser:/root:/bin/sh:0
daemon:*:1:1:0::The devil himself:/:/sbin/nologin:0
```
And on MacOS:
```
_appowner:*:87:87:0::Application Owner:/var/empty:/usr/bin/false:0
_dovenull:*:227:227:0::Dovecot Authentication:/var/empty:/usr/bin/false:0
_netstatistics:*:228:228:0::Network Statistics Daemon:/var/empty:/usr/bin/false:0
```
So despite the similarity of the *BSD manpages, the actual implementation issue appears to be rather FreeBSD-specific.
I don't currently have access to any OpenBSD systems to test, and don't know whether DragonFly is still a thing, but the best for now is to limit the work-around in this PR to just FreeBSD.

**Comment by Bodigrim** [comment]
@vdukhovni would it be possible to split this into three commits, referencing bullet points above? Otherwise it is a bit dense to review. And please update the changelog.

**Comment by hs-viktor** [comment]
> @vdukhovni would it be possible to split this into three commits, referencing bullet points above? Otherwise it is a bit dense to review. And please update the changelog.

Done.

File: System/Posix/User.hsc [hunk 1]
```diff
@@ -67,6 +68,108 @@ import System.IO.Error
...
+--
+-- Consequently, correct portable support for @get(pw|gr)ent(3)@ is rather
+-- non-trivial.  In the threaded runtime, we can run these functions in a
+-- /bound thread/ (via 'forkOS'), thereby avoiding the FreeBSD issues.  We
+-- still need a lock to serialise multiple threads calling these functions
+-- on at last MacOS for lack of @_r@ equivalents.  While on FreeBSD we could
+-- use @getpwent_r(3)@ and @getgrent_r(3)@ in a bound thread without any
+-- locks, implementing this special case is likely not worthwhile.
+--
+-- In the non-threaded runtime, `forkOS` is not available, and so on FreeBSD
```
Bodigrim, line 111: Does modern haddock hyperlink backticked entites as well?

System/Posix/User.hsc
```diff
@@ -67,6 +68,108 @@ import System.IO.Error
...
+--   threads may take turns using the same underlying OS thread, breaking the
+--   thread-safety of the @*_r@ functions, which mutate the file-offset of the
+--   open file shared with any overlapping execution of @*ent(3)@ in the same
+--   thread.
+--
+-- Consequently, correct portable support for @get(pw|gr)ent(3)@ is rather
+-- non-trivial.  In the threaded runtime, we can run these functions in a
+-- /bound thread/ (via 'forkOS'), thereby avoiding the FreeBSD issues.  We
+-- still need a lock to serialise multiple threads calling these functions
+-- on at last MacOS for lack of @_r@ equivalents.  While on FreeBSD we could
```
Bodigrim, line 107: If I understand it correctly,
```suggestion
-- on at least MacOS for lack of @_r@ equivalents.  While on FreeBSD we could
```

System/Posix/User.hsc
```diff
@@ -67,6 +68,108 @@ import System.IO.Error
...
 data {-# CTYPE "struct group"  #-} CGroup
 
+-- -----------------------------------------------------------------------------
+-- Thread safety of passwd/group database access APIs:
+--
+-- All supported unix platforms have @get(pw|gr)(nam|[ug]id)_r(3)@, which
+-- store the result in a caller provided buffer, which solves the most
+-- immediate thread-safety issues.
+--
+-- Things are more complicated for getpwent(3) and getgrent(3).
```
Bodigrim, line 78: Is it feasible to link correspondent man pages and sources for paragraphs below?

System/Posix/User.hsc
hs-viktor: Yes.

System/Posix/User.hsc
hs-viktor: Yes, thanks for the typo fix.

System/Posix/User.hsc
hs-viktor: Actually, looking closely at the FreeBSD code, the issues is actually an unintended side-effect (bug) in the "compat" passwd nsswitch API. The code for compat resets the global iterator key (`st->keynum`) also when doing a [single lookup](https://gitlab.unc.edu.ar/nicolaspapp/freebsd/-/blob/stable/12/lib/libc/gen/getpwent.c#L1774-1783) and then uses it as the [loop control variable](https://gitlab.unc.edu.ar/nicolaspapp/freebsd/-/blob/stable/12/lib/libc/gen/getpwent.c#L1825).

Changing `nsswitch` from `compat` to `files` solves the problem. So this is not a FreeBSD feature, but rather a long-standing bug. I can file a bug report and fix for FreeBSD, but it will still take a while for this to land on user systems. So I think we still need a work-around. :-(

Thoughts?

Bodigrim: Yeah, in the meantime a workaround is needed :(

hs-viktor: This is the fix:
```C
--- lib/libc/gen/getpwent.c     (revision 367119)
+++ lib/libc/gen/getpwent.c     (working copy)
@@ -1726,6 +1726,7 @@
 {
        char                     keybuf[MAXLOGNAME + 1];
        DBT                      key, entry;
+       pwkeynum                 keynum;
        struct compat_state     *st;
        enum nss_lookup_type     how;
        const char              *name;
@@ -1776,9 +1777,10 @@  
                        rv = NS_NOTFOUND;
                        goto fin;
                }
+               keynum = st->keynum;
                stayopen = 1;
        } else {
-               st->keynum = 0;
+               keynum = 0;
                stayopen = st->stayopen;
        }
 docompat:
@@ -1822,13 +1824,13 @@
        }
        key.data = keybuf;
        rv = NS_NOTFOUND;
-       while (st->keynum >= 0) {
-               st->keynum++;
+       while (keynum >= 0) {
+               keynum++;
                if (st->version < _PWD_CURRENT_VERSION) {
-                       memcpy(&keybuf[1], &st->keynum, sizeof(st->keynum));
-                       key.size = sizeof(st->keynum) + 1;
+                       memcpy(&keybuf[1], &keynum, sizeof(keynum));
+                       key.size = sizeof(keynum) + 1;
                } else {
-                       store = htonl(st->keynum);
+                       store = htonl(keynum);
                        memcpy(&keybuf[1], &store, sizeof(store));
                        key.size = sizeof(store) + 1;
                }
@@ -1839,7 +1841,7 @@   
                        rv = NS_UNAVAIL;
                        goto fin;
                } else if (rv == 1) {
-                       st->keynum = -1;
+                       keynum = -1;
                        rv = NS_NOTFOUND;
                        goto fin;
                }
@@ -1924,6 +1926,8 @@   
                        break;
        }
 fin:
+       if (how == nss_lt_all)
+               st->keynum = keynum;
        if (st->db != NULL && !stayopen) {
                (void)st->db->close(st->db);
                st->db = NULL;
```

hs-viktor: I filed a bug report: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=252094
The patch is only for the `getwent()` issue, the `getgrent(3)` issue requires a somewhat more invasive fix, but still looks like a bug.

Bodigrim (approved):

hs-viktor:
@emilypi You'll probably find it easier to review each of the four commits separately, at @Bodigrim's request I split up the changes into separate logical units.

chessai (approved):

chessai:
@hs-viktor I looked through each commit (splitting them up was indeed very helpful), and it looks good to me

vdukhovni:
Rebased on top of new CI.

Bodigrim:
@emilypi this is awaiting for your review.

emilypi:
@Bodigrim thanks for the reminder - I'll review Sunday.

emilypi (approved):

System/Posix/User.hsc
hs-viktor: FWIW, the getrennt(3) issue will also be addressed in a future FreeBSD release. But given the longevity of existing deployments, it'll be some years before we can assume the issue is no longer a concern.

vdukhovni:
For the record, the FreeBSD bug is [fixed](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=252094#c27) in the soon to be released FreeBSD 12.3 and FreeBSD 13.

We could at some point consider whether it is worthwhile to add FreeBSD version-specific CPP guards that revert to standard behaviour on sufficiently recent FreeBSD versions.

Bodigrim:
Nice!

PR #185, opened 2021-05-15
Fixes #81: Added `readDirStreamMaybe`

### LINKED ISSUE #81: Use of empty FilePath to indicate end of DirStream seems deeply questionable
For example:

```
readDirStream :: DirStream -> IO RawFilePath
readDirStream (DirStream dirp) =
  alloca $ \ptr_dEnt  -> loop ptr_dEnt
 where
  loop ptr_dEnt = do
    resetErrno
    r <- c_readdir dirp ptr_dEnt
    if (r == 0)
         then do dEnt <- peek ptr_dEnt
                 if (dEnt == nullPtr)
                    then return BC.empty
                    else do
                     entry <- (d_name dEnt >>= peekFilePath)
                     c_freeDirEnt dEnt
                     return entry
         else do errno <- getErrno
                 if (errno == eINTR) then loop ptr_dEnt else do
                 let (Errno eo) = errno
                 if (eo == 0)
                    then return BC.empty
                    else throwErrno "readDirStream"
```

Wouldn't it be far better if we returned 'Maybe RawFilePath'? Ditto for FilePath API.

mmhat:
This version of `readDirStream` returns a `Maybe FilePath` /
`Maybe ByteString` instead of a (possibly empty) String / ByteString.

Fixes #81

### DIFF
--- a/System/Posix/Directory.hsc
+++ b/System/Posix/Directory.hsc
@@ -31,6 +31,7 @@ module System.Posix.Directory (
    DirStream,
    openDirStream,
    readDirStream,
+   readDirStreamMaybe,
    rewindDirStream,
    closeDirStream,
    DirStreamOffset,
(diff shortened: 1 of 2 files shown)


hs-viktor (approved):
LGTM

PR #186, opened 2021-05-15
Fixes #35: fd{Read,Write} with ByteString payload

### LINKED ISSUE #35: Add convenient I/O operations using `ByteString` for payload
Operations like [`fdRead`](http://hackage.haskell.org/package/unix-2.7.1.0/docs/System-Posix-IO-ByteString.html#v:fdRead) which read the data as `[Char]` are often not what you'd want. Instead you'd want the payload to be a `ByteString`.

Instead of having to roll your own, by using a combination of the `unix-bytestring` and `unix` packages, `unix` ought to directly provide such convenience functions.

See also http://thread.gmane.org/gmane.comp.lang.haskell.libraries/23812/focus=23816

mmhat:
This commit addes two new functions: fdReadBytes and fdWriteBytes.
They are like fdRead and fdWrite respectively but instead of
reading/writing String they use ByteStrings.

Fixes #35

### DIFF
--- a/System/Posix/IO/Common.hsc
+++ b/System/Posix/IO/Common.hsc
@@ -1,7 +1,7 @@
 {-# LANGUAGE CApiFFI #-}
 {-# LANGUAGE NondecreasingIndentation #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -32,6 +32,7 @@ module System.Posix.IO.Common (
     -- EAGAIN exceptions may occur for non-blocking IO!
 
     fdRead, fdWrite,
+    fdReadBytes, fdWriteBytes,
     fdReadBuf, fdWriteBuf,
 
     -- ** Seeking
@@ -60,6 +61,9 @@ module System.Posix.IO.Common (
 
   ) where
 
+import Data.ByteString.Char8 as BC
+import Data.ByteString.Internal as BI
+import Data.ByteString.Unsafe as BU
 import System.IO
 import System.IO.Error
 import System.Posix.Types
@@ -412,6 +416,20 @@ fdRead fd nbytes = do
        s <- peekCStringLen (castPtr buf, fromIntegral n)
        return (s, n)
 
+-- | Read data from an 'Fd'.
+-- Throws an exception if this is an invalid descriptor, or EOF has been
+-- reached.
+fdReadBytes :: Fd
+       -> ByteCount -- ^How many bytes to read
+       -> IO (ByteString, ByteCount) -- ^The bytes read, how many bytes were read.
+fdReadBytes _fd 0 = return (BC.empty, 0)
+fdReadBytes fd nbytes = do
+    BI.createUptoN' (fromIntegral nbytes) $ \ buf -> do
+    rc <- fdReadBuf fd buf nbytes
+    case rc of
+      0 -> ioError (ioeSetErrorString (mkIOError EOF "fdRead" Nothing Nothing) "EOF")
+      n -> return (fromIntegral n, n)
+
 -- | Read data from an 'Fd' into memory.  This is exactly equivalent
 -- to the POSIX @read@ function.
 fdReadBuf :: Fd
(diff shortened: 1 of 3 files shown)


System/Posix/IO/Common.hsc
```diff
@@ -412,6 +416,20 @@ fdRead fd nbytes = do
        s <- peekCStringLen (castPtr buf, fromIntegral n)
        return (s, n)
 
+-- | Read data from an 'Fd'.
+-- Throws an exception if this is an invalid descriptor, or EOF has been
+-- reached.
+fdReadBytes :: Fd
+       -> ByteCount -- ^How many bytes to read
+       -> IO (ByteString, ByteCount) -- ^The bytes read, how many bytes were read.
```
hs-viktor, line 424: `ByteStrings` already carry a length field, there's little point in returning it another copy wrapped in a tuple.  This should use `createUptoN` and return just the `ByteString`.

Also, not all errors are invalid descriptor or `EOF`. On a non-blocking file descriptor, the error could be EAGAIN, and `fdReadBuf` will have thrown the corresponding IO Exception.
I am not sure that raising on `EOF` is the most useful result, returning an empty ByteString may be a cleaner choice, but perhaps we're too used to EOF exceptions in Haskell... :-(

System/Posix/IO/Common.hsc
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE CApiFFI #-}
 {-# LANGUAGE NondecreasingIndentation #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
```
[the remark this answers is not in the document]
mmhat: No, it's not essential. I had to remove the `Safe` pragma because the imported modules of bytestring are not safe.
Likewise modules marked as safe cannot depend on this module anymore, but I am not sure if that regression is a show stopper.

hasufell:
My opinion is that these shouldn't be new functions, but that the existing functions `fdRead`/`fdWrite` in `System.Posix.IO.ByteString` are buggy, since they don't use ByteString. That means they're in fact the same as their counterparts in `System.Posix.IO`:

* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO.html#v:fdRead
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO-ByteString.html#v:fdRead

That makes no sense.

So I think breaking API here is the right choice.

mmhat:
@hasufell I wholeheartedly aggree with you but wanted to be careful with backwards-incompatible changes. If no one objects I'll update this PR such that it fixes this bug and remove the new functions.

hasufell:
@hs-viktor opinions? Users who use the old functions can fix their code easily by importing the ones from System.Posix.IO instead. I think that's a reasonable migration strategy and can likely be automated for large codebases.

System/Posix/IO/Common.hsc
mmhat: > Also, not all errors are invalid descriptor or `EOF`. On a non-blocking file descriptor, the error could be EAGAIN, and `fdReadBuf` will have thrown the corresponding IO Exception.
> I am not sure that raising on `EOF` is the most useful result, returning an empty ByteString may be a cleaner choice, but perhaps we're too used to EOF exceptions in Haskell... :-(

Not sure yet what my opinion is here; I tried to stick to the existing design (with all its bugs and quirks) here. If we decide to change the way how we handle EOF I suggest we do it in https://github.com/haskell/unix/pull/219 and discuss it over there.

mmhat:
> We could also add a test identical to https://github.com/haskell/unix/blob/master/tests/Posix014.hs

@hasufell Will add one once we decide which PR will make it.

Bodigrim:
> That means they're in fact the same as their counterparts in `System.Posix.IO`:
> 
>     * https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO.html#v:fdRead
> 
>     * https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO-ByteString.html#v:fdRead
> 
> 
> That makes no sense.

On the surface level this is kinda expected, the contract is that `System.Posix.ByteString` represents all file paths and environment strings as `ByteString` instead of `String`, but other arguments or outputs remain unchanged. So functions which take neither file paths nor environment strings are indeed the same in both modules. 

https://github.com/haskell/unix/blob/e919c7db25c562651d204bc6d8d3711467708453/System/Posix/ByteString.hs#L16-L21

However, the underlying rationale for the existence of `System.Posix.ByteString` is not performance per se, it's independence from locale encodings. So while it is expected to have some duplicates between `System.Posix` and `System.Posix.ByteString` and it is not a crime to be inefficient, locale-dependent decoding should not have place in `System.Posix.ByteString.fdRead` as it happily happens now. Since there is no way to return `String` without some kind of decoding (locale-driven or other), these functions should indeed return `ByteString`. I think a breaking change is justified here.

hasufell:
Dropped in favor of https://github.com/haskell/unix/pull/219

PR #187, opened 2021-05-15
Fixes #139: Added Show instances in Resource module

### LINKED ISSUE #139: Add Show instances to types in System.Posix.Resource

mmhat:
Fixes #139


Bodigrim (approved):
This looks uncontroversial to me.

hs-viktor (approved):
Might have to revisit this if/when these are changed to use PatternSynonyms, but that may not happen, and can be dealt with when needed.

PR #190, opened 2021-06-06
Disable dependabot

Bodigrim:
I'm tired of dependabot raising a PR for every minor bump of dependencies. We did similar in `bytestring`.


Bodigrim:
CC @vdukhovni

hs-viktor (approved):

PR #191, opened 2021-06-26
derive Read, Show, Eq, and Ord instances for BaudRate

dopamane:
Derive `Read`, `Show`, `Eq`, and `Ord` instances for `BaudRate`. It would be useful to have these instances for [serialport](https://github.com/standardsemiconductor/serialport).

### DIFF
--- a/System/Posix/Terminal/Common.hsc
+++ b/System/Posix/Terminal/Common.hsc
@@ -386,6 +386,7 @@ data BaudRate
   | B3000000
   | B3500000
   | B4000000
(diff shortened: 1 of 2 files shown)


hs-viktor:
I've already handled everything but `Read` in #216. How important is `Read` with symbolic rates? I think that the `Num` instance actually makes for a better UX, and the `BaudRate` constructor is now exposed.

Is #216 sufficient?  I'm inclined to close this if it is...

PR #206, opened 2022-04-14
Update CI

Bodigrim:
Replace FreeBSD build, add CentOS build, i386 build, arm (32 + 64 bits) build.

@hs-viktor this fails on ARM
https://github.com/haskell/unix/blob/9c90c34017f3394076c307932ef900a8abeca67f/tests/Test.hs#L170-L174
with 
```
  posix002:             FAIL (0.04s)
    tests/Test.hs:174:
    expected: "ONE=1\nTWO=2\n"
     but got: "TWO=2\nONE=1\n"
    Use -p '/posix002/' to rerun this test only.
```
Is it acceptable?

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,9 @@
+freebsd_instance:
+  image_family: freebsd-13-0
+
+task:
+  install_script: pkg install -y ghc hs-cabal-install git autoconf
+  script:
+    - cabal update
+    - autoreconf -i
+    - cabal test --test-show-details=direct
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,7 @@
 name: ci
 on:
-  push:
-    branches:
-      - master
-      - 2.7
-  pull_request: {}
+  - push
+  - pull_request
 
 defaults:
   run:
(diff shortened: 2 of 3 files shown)


hs-viktor:
The order of environment variables is not semantically meaningful and so I don't think the test can correctly rely on order the environment variables being preserved by `printenv`.

So we need a better test, that splits the output on newlines and canonicalises the order. Are you willing to fix the test?

hs-viktor:
A related question is whether `executeFile` should *deduplicate* the environment, taking the last instance of each variable.  Presently with `[("ONE", "1"), ("TWO", "2"), ("ONE", "3)]` the environment passed to `printenv` contains both definitions of `ONE`.

Finally, which ARM platform reordered the variables?  I'd expect that perhaps this more an oddity of the `printenv` implementation than of `execute(2)`.  I don't have any ARM machines on hand...

Bodigrim:
> So we need a better test, that splits the output on newlines and canonicalises the order. Are you willing to fix the test?

Yep, fixed. The failure was on Ubuntu, both 32 and 64 bit ARM.

This is ready for review.

hs-viktor (approved):
LGTM.  Do we have anything in the pipeline for the check warnings in `ResuurceLimit.hs`?

Bodigrim:
> Do we have anything in the pipeline for the check warnings in `ResourceLimit.hs`?

I'd probably just disable the warning in `ResourceLimit.hs`: an unexpected constructor will cause the test to file anyway, so it's fine.

PR #207, opened 2022-04-17
Fix the error handling of posix_fallocate

### LINKED ISSUE #156: fileAllocate handles errors incorrectly on some systems
Nice to meet you.

`System.Posix.Fcntl.fileAllocate` appears to handle errors incorrectly. On Linux, `posix_fallocate(3)` states:

> RETURN VALUE
>        posix_fallocate() returns zero on success, or an error number on failure.  Note that errno is not set.

On NetBSD, [`posix_fallocate(2)` has the same behavior](https://netbsd.gw.com/cgi-bin/man-cgi?posix_fallocate+2+NetBSD-7.0).

However, on FreeBSD [`posix_fallocate(2)` returns -1 and sets `errno`](https://www.freebsd.org/cgi/man.cgi?query=posix_fallocate&sektion=2&manpath=FreeBSD+8.3-RELEASE).

`fileAllocate` is [implemented with `throwErrnoIfMinus1_`](https://github.com/haskell/unix/blob/master/System/Posix/Fcntl.hsc#L95) assuming `errno` is set. So it would work on FreeBSD but not on the other OSes.

The correct implementation would be like:

```hs
fileAllocate fd off len = do
  ret <- c_posix_fallocate (fromIntegral fd) (fromIntegral off) (fromIntegral len)
  case () of
    _ | ret == 0 -> pure ()
      -- FreeBSD-style error
      | ret == -1 -> throwErrno "fileAllocate"
      | otherwise -> ioError (errnoToIOError "fileAllocate" (Errno ret) Nothing Nothing)
```

or some `#if/#else` guards rather than this hack?

matil019:
Resolves #156 

~~This commit introduces a CPP guard to take care of the difference between OSes of the errors of `posix_fallocate`.~~

**Update:** No conditionals are added, just the error handler is updated to correctly handle the error value returned by `posix_fallocate`.

### DIFF
--- a/System/Posix/Fcntl.hsc
+++ b/System/Posix/Fcntl.hsc
@@ -87,16 +87,11 @@ fileAdvise _ _ _ _ = return ()
 -- @since 2.7.1.0
 fileAllocate :: Fd -> FileOffset -> FileOffset -> IO ()
 #if HAVE_POSIX_FALLOCATE
-#if defined(freebsd_HOST_OS)
-fileAllocate fd off len = do
-  throwErrnoIfMinus1_ "fileAllocate" (c_posix_fallocate (fromIntegral fd) (fromIntegral off) (fromIntegral len))
-#else
 fileAllocate fd off len = do
   ret <- c_posix_fallocate (fromIntegral fd) (fromIntegral off) (fromIntegral len)
   if ret == 0
     then pure ()
     else ioError (errnoToIOError "fileAllocate" (Errno ret) Nothing Nothing)
-#endif
 
 foreign import capi safe "fcntl.h posix_fallocate"
   c_posix_fallocate :: CInt -> COff -> COff -> IO CInt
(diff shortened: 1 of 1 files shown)


System/Posix/Fcntl.hsc
```diff
@@ -87,8 +87,16 @@ fileAdvise _ _ _ _ = return ()
 -- @since 2.7.1.0
 fileAllocate :: Fd -> FileOffset -> FileOffset -> IO ()
 #if HAVE_POSIX_FALLOCATE
+#if defined(freebsd_HOST_OS)
 fileAllocate fd off len = do
   throwErrnoIfMinus1_ "fileAllocate" (c_posix_fallocate (fromIntegral fd) (fromIntegral off) (fromIntegral len))
+#else
```
vdukhovni, line 93: The FreeBSD exception is obsolete as of FreeBSD 11.0. Just go with the same code on all platforms that have `posix_fallocate(2)`.

matil019: Done.

hs-viktor:
CI failed, but I just merged a CI fix PR. Perhaps rebasing will resolve the issue.  Please rebase.

matil019:
Rebased.

hs-viktor:
Forgot about the haskell.org outage, will have to retry the CI later...

Bodigrim:
CI is all green.

Bodigrim (approved):

hs-viktor (approved):

PR #203, opened 2022-03-30
Update autoconf to 2.71

bgamari:

PR #208, opened 2022-04-17
Update autoconf to 2.69

Bodigrim:
Supersedes #203 with a more conservative lower bound for autoconf.


Bodigrim:
CI demonstrates that it's better to retain `config.sub`.

hs-viktor (approved):
LGTM

hs-viktor:
FWIW, Autoconf 2.69 was released [10 years ago](https://lists.gnu.org/archive/html/autotools-announce/2012-04/msg00004.html) so should be a fairly conservative floor.

PR #211, opened 2022-05-03
Remove buildable check based on dlfcn.h

TerrorJack:
This PR is a predecessor to #205. We shall no longer deny building for platforms lacking `dlfcn.h`, but rely on fine-grained checks on individual interfaces instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,10 +11,6 @@ AC_USE_SYSTEM_EXTENSIONS
 
 AC_CONFIG_HEADERS([include/HsUnixConfig.h])
 
-# Is this a Unix system?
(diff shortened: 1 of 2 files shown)


TerrorJack:
Hi, would you take another look at this PR? The aarch64 CI failure is irrelevant and seem to be random timeout, shouldn't be a problem if the 60min limit is bumped a bit.

hs-viktor:
I am not sure why this makes sense as a stand-alone PR.  The `dlfcn.h` header is still included in `HsUnix.h`, so the `unix` package won't build if it is missing without further changes.  Why not make this change in a PR that actually makes `unix` not unconditionally depend on `dlfcn.h`.

TerrorJack:
And that PR would be #205. @Bodigrim indeed the changes here are too small to be meaningful, would you agree on closing this one and getting back to #205 once the ghc warnings are fixed there?

PR #212, opened 2022-05-07
Don't show repo as modified after configure

shlevy:


hs-viktor (approved):

gergoerdi:
I believe this should also have removed `config.sub` from the repo, instead of just adding it to `.gitignore`. See e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/19574#note_441377

PR #213, opened 2022-05-07
Don't show repo as modified after configure

shlevy:


shlevy:
Note that this is the same commit as #212, no need to cherry-pick etc. if merged with a proper merge commit.

hs-viktor (approved):

PR #74, opened 2016-10-07
Support other baud rates

mfine:
We're using higher baud rates not defined - we'd like to access them if our underlying system can support them. The limited set of baud rates is currently too restrictive - this allows an escape hatch.

PR #216, opened 2022-05-30
Pattern baudrate take two

hs-viktor:
This should subsume and polish #145 (which I will close along with #74)

### DIFF
--- a/System/Posix/Terminal/Common.hsc
+++ b/System/Posix/Terminal/Common.hsc
@@ -1,4 +1,6 @@
 {-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE Trustworthy #-}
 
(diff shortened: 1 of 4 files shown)


System/Posix/Terminal/ByteString.hsc
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE PatternSynonyms #-}
```
hs-viktor, line 2: Actually, this pragma is not needed here, the synonyms are bundled in the defining module...  I'll revert this after CI passes, or if a fixup is needed to get it to pass.

hs-viktor: Done.

hs-viktor:
Removed `#ifdefs` from the POSIX-mandated speeds, and extended/added haddock comments.

Bodigrim (approved):

Bodigrim:
Looks great, please go ahead with a merge.

PR #219, opened 2022-07-05
Fixes #35: fd{Read,Write} with ByteString payload (Breaking change)

### LINKED ISSUE #35: Add convenient I/O operations using `ByteString` for payload

mmhat:
Version of https://github.com/haskell/unix/pull/186 with API breakage. Feel free to close this one or the other one.

### DIFF
--- a/System/Posix/IO.hsc
+++ b/System/Posix/IO.hsc
@@ -63,6 +63,13 @@ module System.Posix.IO (
 
   ) where
 
+import Foreign ( allocaBytes, castPtr )
+import Foreign.C ( peekCStringLen, withCStringLen )
+
+import GHC.IO.Exception ( IOErrorType(EOF) )
+
+import System.IO.Error ( ioeSetErrorString, mkIOError )
+
 import System.Posix.Types
 import System.Posix.Error
 import System.Posix.IO.Common
@@ -88,9 +95,9 @@ openFdAt :: Maybe Fd -- ^ Optional directory file descriptor
          -> OpenFileFlags -- ^ Append, exclusive, truncate, etc.
          -> IO Fd
 openFdAt fdMay name how flags =
-   withFilePath name $ \str ->
-     throwErrnoPathIfMinus1Retry "openFdAt" name $
-       openat_ fdMay str how flags
+  withFilePath name $ \str ->
+    throwErrnoPathIfMinus1Retry "openFdAt" name $
+      openat_ fdMay str how flags
 
 -- |Create and open this file in WriteOnly mode.  A special case of
 -- 'openFd'.  See 'System.Posix.Files' for information on how to use
@@ -111,3 +118,25 @@ createFileAt :: Maybe Fd -- ^ Optional directory file descriptor
              -> IO Fd
 createFileAt fdMay name mode
   = openFdAt fdMay name WriteOnly defaultFileFlags{ trunc=True, creat=(Just mode) }
+
+-- | Read data from an 'Fd' and convert it to a 'String' using the locale encoding.
+-- Throws an exception if this is an invalid descriptor, or EOF has been
+-- reached.
+fdRead :: Fd
+       -> ByteCount -- ^How many bytes to read
+       -> IO (String, ByteCount) -- ^The bytes read, how many bytes were read.
+fdRead _fd 0 = return ("", 0)
+fdRead fd nbytes =
+  allocaBytes (fromIntegral nbytes) $ \ buf -> do
+  rc <- fdReadBuf fd buf nbytes
+  case rc of
+    0 -> ioError (ioeSetErrorString (mkIOError EOF "fdRead" Nothing Nothing) "EOF")
+    n -> do
+      s <- peekCStringLen (castPtr buf, fromIntegral n)
+      return (s, n)
+
+-- | Write a 'String' to an 'Fd' using the locale encoding.
+fdWrite :: Fd -> String -> IO ByteCount
+fdWrite fd str =
+  withCStringLen str $ \ (buf,len) ->
+    fdWriteBuf fd (castPtr buf) (fromIntegral len)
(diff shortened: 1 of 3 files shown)


Bodigrim:
Could someone review: is it the only place where `*.ByteString` modules use `String`?

hasufell:
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO-ByteString.html#v:fdRead
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-IO-ByteString.html#v:fdWrite
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-DynamicLinker-Module-ByteString.html
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-DynamicLinker-ByteString.html
* https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-ByteString-FilePath.html (not very important probably)

I think only `fdRead` and `fdWrite` are interesting. The other issue is that there's no bytestring variant of https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-User.html ...wich includes two types `groupEntry` and `UserEntry`, which have `String` records.

hasufell:
I think this PR is fine as-is and doesn't need a wider scope.

@Bodigrim @hs-viktor let's vote on this issue:

1. accept this PR
2. accept PR https://github.com/haskell/unix/pull/186
3. accept none of them

----

I clearly vote for 1. :+1:

System/Posix/IO.hsc
```diff
@@ -111,3 +118,25 @@ createFileAt :: Maybe Fd -- ^ Optional directory file descriptor
              -> IO Fd
 createFileAt fdMay name mode
   = openFdAt fdMay name WriteOnly defaultFileFlags{ trunc=True, creat=(Just mode) }
+
+-- | Read data from an 'Fd' and convert it to a 'String' using the locale encoding.
+-- Throws an exception if this is an invalid descriptor, or EOF has been
+-- reached.
+fdRead :: Fd
```
hs-viktor, line 125: It may be worth noting that unless the input is known to be ASCII data, or an 8-bit locale is in use, reading `n` bytes and expecting a `String` is rather dubious.  One can read and decode *lines* (in an UTF-8 locale) because UTF-8 is self-synchronising, and LF or CRLF is never in the middle of a multi-byte sequence.  Otherwise, reading "n-bytes" from a file into a string is a recipe for failure.  Use of this function should be discouraged.

hs-viktor (commented):
If we're introducing an API break, we could also *deprecate* the `String` variants, they are semantically dubious.

hs-viktor:
I'm tempted to accept this PR, but have no idea how much breakage would ensue downstream.  How widely used are these functions?

Bodigrim:
I think this is an acceptable level of breakage. The usage is quite limited, e. g., https://hackage-search.serokell.io/?q=%5CbfdRead%5Cb - and most of these are for non-ByteString version.

I vote to accept.

hs-viktor:
I will accept this PR, modulo additional documentation discouraging the `String` variants, if not outright deprecating them.

System/Posix/IO.hsc
hasufell: I'm not sure I follow.

This issues seems to be present everywhere, including in the handling of filepaths. It's broken: https://gist.github.com/hasufell/c600d318bdbe010a7841cc351c835f92

Anything that uses `getLocaleEncoding`, `getFileSystemEncoding` or `getForeignEncoding` to convert `CString` to `String` potentially runs into bugs, no?

Following that, we'd need to deprecate half of base and most of the unix String based API.

hs-viktor: > Anything that uses `getLocaleEncoding`, `getFileSystemEncoding` or `getForeignEncoding` to convert `CString` to `String` potentially runs into bugs, no?

No, the issue here is that we're not reading a *complete* string, we're reading `n bytes` from a file.  These `n bytes` may well end in the middle of a multi-byte grapheme.

Converting *entire* file names or similar octet data (rather than `n byte` fragments of strings) to Strings is much safer under reasonable assumptions.

hasufell: > These n bytes may well end in the middle of a multi-byte grapheme.

Right, got it. And when you append the strings, you get garbage.

Bodigrim:
Let's deprecate `String` variants as unsafe then. @mmhat could you please add this to the PR?

System/Posix/IO.hsc
hs-viktor: > > These n bytes may well end in the middle of a multi-byte grapheme.
> 
> Right, got it. And when you append the strings, you get garbage.

Yes, the description of the action says it all: *read n bytes*.  That's a `ByteString` operation, not a `String` operation.  `String`s don't consist of bytes.

hasufell:
> Let's deprecate `String` variants as unsafe then. @mmhat could you please add this to the PR?

I can make a follow-up PR. I'd argue the scope of this PR is just about the ByteString variants.

hasufell (approved):

Bodigrim (approved):

hs-viktor:
The String variants do of course work as expected when the handle is in "binary mode". The more cautious alternative to deprecation is caveats in the haddocks.

Issue #197, opened 2021-11-28
How to convert ClockTick to time difference?


yaitskov:
Hi,

getProcessTimes returns times in ClockTick.
How to convert values to human readable format? i.e. seconds.
I noticed that CClock is defined in base library, but there is not much haddock about it.
hoogle cannot find (:: CClock -> NormalDiffTime)

PR #221, opened 2022-07-10
Add clocksPerSec wrt #197

hasufell:


System/Posix/Process/Common.hsc
```diff
@@ -302,13 +303,31 @@ data {-# CTYPE "struct tms" #-} CTms
 foreign import capi unsafe "HsUnix.h times"
   c_times :: Ptr CTms -> IO CClock
 
+-- | Returns the value from the @CLOCK_PER_SEC@ macro, which is required by POSIX.
+clocksPerSec :: ClockTick
+#ifdef HAVE_CLOCKS_PER_SEC
+clocksPerSec = unsafePerformIO c_clocks_per_sec
+
+foreign import capi unsafe "HsUnix.h __hsunix_clocks_per_second"
+  c_clocks_per_sec :: IO CClock
```
hs-viktor, line 312: Why is this an `IO` action at all.  It can't sensibly vary, without making seconds since boot, or intervals that cross the change undefined. A pure function should suffice here, since in particular we're calling unsafePerformIO to evaluate it once anyway.

hs-viktor (changes requested):

Issue #4, opened 2014-07-28
handleToFd leaks memory


ghost:
On unix-2.6.0.1 with GHC 7.6.3 the following program leaks memory very fast:

``` haskell
import Control.Monad
import System.IO
import System.IO.Error
import System.Posix.IO
import System.Posix.Types

test :: IO ()
test = do
  let path = "/tmp/leak-test"
  handle <- openFile path WriteMode
  putStrLn $ "Opened file " ++ path
  fd <- handleToFd handle
  putStrLn $ "Converted to handle " ++ path
  _ <- tryIOError $ closeFd fd
  putStrLn $ "Closed file " ++ path

main = forever test
```

This is only due to `handleToFd`. Opening a file directly using `openFd` doesn't leak.

rwbarton:
This seems to be some issue with broken locales related to GHC Trac #7695 and #10298. I can reproduce it on an ordinary system after `export LC_ALL=badlc`. I'd guess the allocations are happening in iconv.

argiopetech:
Nice find. It doesn't leak very quickly, but this definitely triggers the issue on my machine.

rwbarton:
OK, the cause is rather simple. Compare the eventual implementation of `hClose` from base
```
hClose_handle_ :: Handle__ -> IO (Handle__, Maybe SomeException)
hClose_handle_ h_@Handle__{..} = do

    -- close the file descriptor, but not when this is the read
    -- side of a duplex handle.
    -- If an exception is raised by the close(), we want to continue
    -- to close the handle and release the lock if it has one, then
    -- we return the exception to the caller of hClose_help which can
    -- raise it if necessary.
    maybe_exception <-
      case haOtherSide of
        Nothing -> trymaybe $ IODevice.close haDevice
        Just _  -> return Nothing

    -- free the spare buffers
    writeIORef haBuffers BufferListNil
    writeIORef haCharBuffer noCharBuffer
    writeIORef haByteBuffer noByteBuffer

    -- release our encoder/decoder
    closeTextCodecs h_

    -- we must set the fd to -1, because the finalizer is going
    -- to run eventually and try to close/unlock it.
    -- ToDo: necessary?  the handle will be marked ClosedHandle
    -- XXX GHC won't let us use record update here, hence wildcards
    return (Handle__{ haType = ClosedHandle, .. }, maybe_exception)
```
to that of `handleToFd` from unix
```
handleToFd' :: Handle -> Handle__ -> IO (Handle__, Fd)
[...]

PR #222, opened 2022-07-10
Fix memory leak in handleToFd, fixes issue #4

hasufell:
From comment https://github.com/haskell/unix/issues/4#issuecomment-123133365

I'm not really sure this is the correct thing to do.

@rwbarton @hs-viktor

### DIFF
--- a/System/Posix/IO/Common.hsc
+++ b/System/Posix/IO/Common.hsc
@@ -1,7 +1,7 @@
 {-# LANGUAGE CApiFFI #-}
 {-# LANGUAGE NondecreasingIndentation #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -73,7 +73,10 @@ import GHC.IO.Handle.Types
 import qualified GHC.IO.FD as FD
 import qualified GHC.IO.Handle.FD as FD
 import GHC.IO.Exception
+import GHC.IO.Buffer
+import GHC.IORef
 import Data.Typeable (cast)
+import System.IO.Unsafe (unsafePerformIO)
 
 #if !defined(HAVE_PIPE)
 import System.IO.Error ( ioeSetLocation )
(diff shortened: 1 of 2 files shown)


Bodigrim:
You probably want to replace `Safe` with `Trustworthy`, otherwise downstream consumers will complain. 
(I personally have zero sympathy to Safe Haskell, but such changes do break other packages in a very frustrating way)

System/Posix/IO/Common.hsc
```diff
@@ -301,10 +304,26 @@ handleToFd' h h_@Handle__{haType=_,..} = do
      -- converting a Handle into an Fd effectively means
      -- letting go of the Handle; it is put into a closed
      -- state as a result.
-     flushWriteBuffer h_
+
+     -- free the spare buffers
+     writeIORef haBuffers BufferListNil
+     writeIORef haCharBuffer noCharBuffer
+     writeIORef haByteBuffer noByteBuffer
```
hs-viktor, line 311: Doesn't this change lose any unwritten data in the write buffer?  It used to be flushed, but now is just dropped?

With `hClose` this happens before the call to `hClose_handle_ handle_` (after which the above code is modelled):
```
hClose_help :: Handle__ -> IO (Handle__, Maybe SomeException)
hClose_help handle_ =
  case haType handle_ of
      ClosedHandle -> return (handle_,Nothing)
      _ -> do mb_exc1 <- trymaybe $ flushWriteBuffer handle_ -- interruptible
                    -- it is important that hClose doesn't fail and
                    -- leave the Handle open (#3128), so we catch
                    -- exceptions when flushing the buffer.
              (h_, mb_exc2) <- hClose_handle_ handle_
              return (h_, if isJust mb_exc1 then mb_exc1 else mb_exc2)
```

hs-viktor (changes requested):

System/Posix/IO/Common.hsc
hasufell: I'm really not too sure about this change. I think it's not necessary for 2.8.0.0, since people have lived with it for over a decade.

hs-viktor: If the `memory leak` can be safely avoided, I'm inclined to fix anyway, but replacing flush with just dropping the buffer is not correct.  Otherwise, I don't see any issues.

hasufell: @hs-viktor I'm not too familiar with this part of the code. Feel free to adjust this PR.

hasufell:
Btw, I tried to reproduce the original bug and the only GHC I could reproduce it with is **8.0.2** (I didn't try olders).

But with 8.0.2 is builds up pretty fast.

hasufell:
Since current unix cannot be built with 8.0.2 I suggest to just close this as obsolete.

hasufell:
Or we try to fix it for `2.7.2.3`, but it won't be shipped with any affected GHC.

Bodigrim:
I don't see any point in `unix-2.7.2.3`, let's just close the issue as non-reproducible.

PR #223, opened 2022-07-10
Expose DirStream and friends, fixes #25

### LINKED ISSUE #25: export System.Posix.Directory.Common
Without `DirStream`, if you wish to do a deep traversal of a directory structure, it's necessary to stat everything in order to determine if it's a sub-directory (and needs to be descended into) or a regular file.  This imposes a significant performance penalty (see https://github.com/JohnLato/posix-paths for benchmarks).

I don't think there's any reason this module can't be exported in its entirety.

hasufell:
@JohnLato

### DIFF
--- a/System/Posix/Directory/Internals.hsc
+++ b/System/Posix/Directory/Internals.hsc
@@ -1,0 +1,17 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  System.Posix.Directory.Internals
+-- Copyright   :  (c) The University of Glasgow 2022
+-- License     :  BSD-style (see the file libraries/base/LICENSE)
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  provisional
+-- Portability :  non-portable (requires POSIX)
+--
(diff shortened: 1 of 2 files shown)


Bodigrim:
The prior art suggests `*.Internals`, not `*.Internal`.
The issues looks like it would be sufficient to export data types only. In this case it could be `*.Types`.

hs-viktor:
> The prior art suggests `*.Internals`, not `*.Internal`.
> The issues looks like it would be sufficient to export data types only. In this case it could be `*.Types`.

I don't have a horse in this race, there are plenty of packages that export `.Internal` (e.g. `bytestring`).  No objections provided the user knows they're dabbling in interfaces outside the stability contract.

Bodigrim:
I meant "prior art" in this very same package, [`System.Posix.Process.Internals`](https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-Process-Internals.html).

hasufell:
Done

Bodigrim (approved):

hs-viktor (approved):

PR #226, opened 2022-07-11
Fix wasm CI / disable for PRs

hasufell:
@TerrorJack 

This is starting to get a little cumbersome. I'm inclined to make the wasm job non-fatal until it's straightened out. It's blocking PRs that have nothing to do with it.

### DIFF
--- a/cabal.project.wasm32-wasi
+++ b/cabal.project.wasm32-wasi
@@ -4,3 +4,5 @@ package unix
   ghc-options: -Wno-unused-imports
 
 write-ghc-environment-files: always
+
+allow-newer: all:base


hasufell:
I want the wasm job disabled for PRs for now. Otherwise I'm not making any progress here.

hasufell:
It seems it's working now...

@TerrorJack my guess is that these things should work better on GHC gitlab CI. Migrating there would also allow us to mark jobs as non-fatal. Github doesn't have this feature yet. Opinions?

TerrorJack:
It was a flakiness in head.hackage server, since the setup script of `ghc-wasm32-wasi` will perform a cabal update, which fetches from both hackage and head.hackage. I'm not sure if running those in ghc gitlab ci will be more robust, given ghc gitlab ci's track record of robustness.

I suggest only pushing the `allow-newer: base` change because that allows wasm32 CI job to continue to function, without blocking any ongoing PR. If you believe wasm32 CI job should still be disabled, please let me know and I'll open an alternative PR that moves the logic to a bash script that never fails, so jobs are always green and I'll manually check the status of each single actions run in the repo and offer fixes when needed.

hasufell:
> I'm not sure if running those in ghc gitlab ci will be more robust, given ghc gitlab ci's track record of robustness.

We have a full-time devops working on it. We're relying on it already for ghcup, filepath, cabal (releases) and HLS (releases).

Bodigrim:
I'm not a fan of moving `unix` to GitLab. I'm already spreaded too thin and do not want to split attention between two platforms. 

If `allow-newer: all:base` is sufficient, I'd prefer to keep WASM job for pull requests. Otherwise it is too easy to forget about it.

chessai (approved):

PR #228, opened 2022-07-16
Synchronize ByteString and String modules

### LINKED ISSUE #68: String and ByteString implementations have desynchronized
I've been Backpack'ifying unix on the suggestion of @hvr, and I have realized that the String and ByteString implementations have desynchronized in a few cases. I'm still working my way through the modules but here are the ones I've noticed:

System.Posix.Env
- ByteString does not implement/export setEnvironment
- ByteString: `unsetEnv name = putEnv (name ++ "=")` Pretty sure this is a type error
- ByteString: `putEnv` uses `useAsCString` to allocate the string passed to `putenv`, but String has a comment indicating we must not free the string after putenv according to SUSv2

Actually, amazingly enough, I checked the rest of the implementations and this was the only desynchronization.

hasufell:
And fix free-bug in 'putEnv'. Fixes #68

### DIFF
--- a/System/Posix/Env/ByteString.hsc
+++ b/System/Posix/Env/ByteString.hsc
@@ -1,5 +1,5 @@
 {-# LANGUAGE CApiFFI #-}
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -21,9 +21,11 @@ module System.Posix.Env.ByteString (
         , getEnvDefault
         , getEnvironmentPrim
         , getEnvironment
+        , setEnvironment
         , putEnv
         , setEnv
-       , unsetEnv
+        , unsetEnv
+        , clearEnv
 
        -- * Program arguments
        , getArgs
@@ -31,13 +33,16 @@ module System.Posix.Env.ByteString (
 
 #include "HsUnix.h"
 
+import Control.Monad
 import Foreign
 import Foreign.C
 import Data.Maybe       ( fromMaybe )
 
+import System.Posix.Env ( clearEnv )
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Char8 as BC
 import Data.ByteString (ByteString)
+import Data.ByteString.Internal (ByteString (PS), memcpy)
 
 -- |'getEnv' looks up a variable in the environment.
 
@@ -96,6 +101,18 @@ getEnvironment = do
       | BC.head y == '=' = (x,B.tail y)
       | otherwise       = error $ "getEnvironment: insane variable " ++ BC.unpack x
 
+-- |'setEnvironment' resets the entire environment to the given list of
+-- @(key,value)@ pairs.
+--
+-- @since 2.8.0.0
+setEnvironment ::
+  [(ByteString,ByteString)] {- ^ @[(key,value)]@ -} ->
+  IO ()
+setEnvironment env = do
+  clearEnv
+  forM_ env $ \(key,value) ->
+    setEnv key value True {-overwrite-}
+
 -- |The 'unsetEnv' function deletes all instances of the variable name
 -- from the environment.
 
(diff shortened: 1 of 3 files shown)


System/Posix/Env/ByteString.hsc
```diff
@@ -116,15 +135,19 @@ foreign import capi unsafe "HsUnix.h unsetenv"
    c_unsetenv :: CString -> IO ()
 # endif
 #else
-unsetEnv name = putEnv (name ++ "=")
+unsetEnv name = putEnv (name <> BC.pack "=")
```
hasufell, line 138: This is in fact a compile failure that was never catched

System/Posix/Env/ByteString.hsc
```diff
@@ -116,15 +135,19 @@ foreign import capi unsafe "HsUnix.h unsetenv"
...
 
 putEnv :: ByteString {- ^ "key=value" -} -> IO ()
-putEnv keyvalue = B.useAsCString keyvalue $ \s ->
-  throwErrnoIfMinus1_ "putenv" (c_putenv s)
+putEnv (PS fp _ l) =
+  allocaBytes (l+1) $ \buf -> do
+    let p = unsafeForeignPtrToPtr fp
+    memcpy buf p l
+    pokeByteOff buf l (0::Word8)
+    throwErrnoIfMinus1_ "putenv" (c_putenv (castPtr buf))
```
hasufell, line 150: This fixes

```
Prelude> import System.Posix.Env.ByteString
Prelude System.Posix.Env.ByteString> :set -XOverloadedStrings
Prelude System.Posix.Env.ByteString> putEnv "foo=bar"
Prelude System.Posix.Env.ByteString> getEnv "foo"
Just "bar"
Prelude System.Posix.Env.ByteString> getEnv "foo"
Nothing
```

oops...

System/Posix/Env/ByteString.hsc
```diff
@@ -96,6 +103,18 @@ getEnvironment = do
       | BC.head y == '=' = (x,B.tail y)
       | otherwise       = error $ "getEnvironment: insane variable " ++ BC.unpack x
 
+-- |'setEnvironment' resets the entire environment to the given list of
+-- @(key,value)@ pairs.
+--
+-- @since 2.8.0.0
+setEnvironment ::
```
hasufell, line 110: It's present in the String module

System/Posix/Env/ByteString.hsc
```diff
@@ -116,15 +134,25 @@ foreign import capi unsafe "HsUnix.h unsetenv"
...
+  --
+  -- "the string pointed to by string shall become part of the environment,
+  -- so altering the string shall change the environment. The space used by
+  -- string is no longer used once a new string which defines name is passed to putenv()."
+  --
+  -- hence we must not free the buffer
+  buf <- mallocBytes (l+1)
+  memcpy buf (unsafeForeignPtrToPtr fp) l
+  pokeByteOff buf l (0::Word8)
+  throwErrnoIfMinus1_ "putenv" (c_putenv (castPtr buf))
```
hasufell, line 155: Old code:

```
Prelude> import System.Posix.ByteString
Prelude System.Posix.ByteString> :set -XOverloadedStrings
Prelude System.Posix.ByteString> getEnv "foo"
Nothing
Prelude System.Posix.ByteString> putEnv "foo=bar"
Prelude System.Posix.ByteString> getEnv "foo"
Just "bar"
Prelude System.Posix.ByteString> getEnv "foo"
Nothing
```

System/Posix/Env/ByteString.hsc
```diff
@@ -116,15 +133,25 @@ foreign import capi unsafe "HsUnix.h unsetenv"
...
+unsetEnv name = putEnv (name <> BC.pack "=")
 #endif
 
 -- |'putEnv' function takes an argument of the form @name=value@
 -- and is equivalent to @setEnv(key,value,True{-overwrite-})@.
 
 putEnv :: ByteString {- ^ "key=value" -} -> IO ()
-putEnv keyvalue = B.useAsCString keyvalue $ \s ->
-  throwErrnoIfMinus1_ "putenv" (c_putenv s)
+putEnv (PS fp _ l) = withForeignPtr fp $ \p -> do
```
hs-viktor, line 143: Is this code really going to be built against ancient `bytestring` versions that have `PS` instead of `BS`?  And if it were, it would be *very wrong* to ignore the offset.  Use `BS` instead.

Bodigrim: Using `BS` would require `build-depends: bytestring >= 0.11`. That's acceptable, but it's easy to be more conservative with 
```haskell
putEnv (PS fp o l) = withForeignPtr fp $ \p -> do ... (p `plusPtr` o)
```

System/Posix/Env/ByteString.hsc
Bodigrim:
```suggestion
unsetEnv name = putEnv (BC.snoc name '=')
```

Bodigrim (commented):
Thanks for fixing this @hasufell, the bug looked grave. Would it be possible to add a regression test?

System/Posix/Env/ByteString.hsc
hasufell: Sorry... it was almost 2 am at night.

tests/PutEnv001.hs
```diff
@@ -0,0 +1,32 @@
...
+  print env
+  env <- System.Posix.Env.ByteString.getEnv (fromString "foo")
+  performMinorGC
+  print env
+  env <- System.Posix.Env.ByteString.getEnv (fromString "foo")
+  performMinorGC
+  print env
+  env <- System.Posix.Env.ByteString.getEnv (fromString "foo")
+  print env
+  env @?= Just (fromString "bar")
```
hasufell, line 32: The failure is a little hard to trigger... but this worked for me and the `print`s are necessary.

Bodigrim (approved):

Issue #62, opened 2016-05-19
forkProcess fails with +RTS -N


jbaum98:
I have a program in which I use `forkProcess`. When it is run with `-threaded +RTS -N1`, it works fine, but when I use `-threaded +RTS -N2` I get this error:
```
failed to create OS thread: Resource temporarily unavailable
```
I haven't been able to reproduce this issue in a small test case, only within the larger program. To provide some context though, it' called only from the following function:
```haskell
pOpen :: FilePath -> [String] -> (Handle -> IO a) -> IO a
pOpen fp args func = do
  pipepair <- createPipe
  pid <- do
    p <- Control.Exception.try (forkProcess $ childstuff pipepair)
    case p of
      Right x -> return x
      Left (e :: Control.Exception.IOException) -> fail ("Error in fork: " ++ show e)
  retval <- callfunc pipepair
  let rv = seq retval retval
  void $ getProcessStatus True False pid
  return rv
  where
    callfunc pipepair = do
      closeFd (fst pipepair)
      h <- fdToHandle (snd pipepair)
      x <- func h
      hClose h
      return $! x

    childstuff pipepair = do
              void $ dupTo (fst pipepair) stdInput
              closeFd $ fst pipepair
              closeFd $ snd pipepair
              executeFile fp True args Nothing
```
GHC Version: 7.10.3
unix version: 2.7.1.0
OS: Ubuntu 16.04

jbaum98:
In versions around 2.4, it mentioned this bug in the documentation. For example in this commit: 9f0c96ed1eb9eda5e74f13acf1ddc3ab523b0bd4
But later on in this commit (9f0c96ed1eb9eda5e74f13acf1ddc3ab523b0bd4) that warning was removed.

xtendo-org:
I confirm that this indeed happens.

Running programs that contain `forkProcess` with `+RTS -N2` causes:

```
failed to create OS thread: Resource temporarily unavailable
```

This is not deterministic; sometimes it happens, sometimes it doesn't.

As mentioned in 9f0c96e, shared resources also stumble. I try to create pipes with `createPipe`, but if I use `dupTo` to these pipe FDs in the child process, I occasionally get:

```
dupTo: resource busy (Device or resource busy)
```

I'd be very grateful if anyone can look into this.

hvr:
since it was @simonmar who removed that comment, he may know more about...

simonmar:
[...]

PR #231, opened 2022-07-17
Improve 'forkProcess' documentation wrt #62

hasufell:
See #62 

This is not a proper fix. Not sure there is one.


System/Posix/Process/Common.hsc
```diff
@@ -449,6 +449,10 @@ The exception masking state of the executed action is inherited
 threads are not copied into the child process, it's easy to go wrong:
 e.g. by accessing some shared resource that was held by another thread
 in the parent.
+
+GHC note: 'forkProcess' is not currently very well supported when using multiple
+processors (@+RTS -N@), although it is supported with @-threaded@ as
+long as only one processor is being used.
```
hs-viktor, line 455: I think `processor` is the wrong aspect of the issue.  A program can be multi-threaded even on a single CPU.  This is really is about (OS) threads. If the RTS is running in a single-threaded process, all is well.  In multi-threaded processes, fork(2) and friends have poor semantics. A better solution to the OP's problem would be `posix_spawn`.

hasufell: I just copy pasted the original commit from unix: https://github.com/haskell/unix/commit/05eea1ea715745d4e2086d3b25a14f35f424045c

What exact wording do you propose?

hs-viktor: I would replace processor with either __OS thread__ or __capability__

PR #238, opened 2022-07-22
Avoid redundant pattern warning in Resource.hsc

sheaf:
With [GHC MR !8478](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8478), GHC is able to spot a redundant pattern match
when RLIM_SAVED_CUR == RLIM_SAVED_MAX, which it wasn't able to
detect before. So we disable -Woverlapping-patterns for this module.
The alternatives seem more unsatisfactory:

  - we can't perform the equality test RLIM_SAVED_CUR == RLIM_SAVED_MAX
    using CPP macros, because one of the values might expand out to
    have casts;
  - using 'considerAccessible' causes an issue with the Safe Haskell
    annotation, as GHC.Exts isn't Safe;
  - adding a dummy equation such as "id True" to the first pattern match
    would work, but seems more ad-hoc.

### DIFF
--- a/System/Posix/Resource.hsc
+++ b/System/Posix/Resource.hsc
@@ -1,9 +1,6 @@
 {-# LANGUAGE CApiFFI #-}
-#if __GLASGOW_HASKELL__ >= 709
-{-# LANGUAGE Safe #-}
-#else
 {-# LANGUAGE Trustworthy #-}
-#endif
+
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  System.Posix.Resource
@@ -30,6 +27,10 @@ module System.Posix.Resource (
 import System.Posix.Types
 import Foreign
 import Foreign.C
+
+#if __GLASGOW_HASKELL__ >= 905
+import GHC.Exts ( considerAccessible )
+#endif
 
 -- -----------------------------------------------------------------------------
 -- Resource limits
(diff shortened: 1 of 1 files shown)


System/Posix/Resource.hsc
```diff
@@ -100,12 +103,16 @@ unpackRLimit :: CRLim -> ResourceLimit
 unpackRLimit (#const RLIM_INFINITY)  = ResourceLimitInfinity
 unpackRLimit other
 #if defined(RLIM_SAVED_MAX)
-    | ((#const RLIM_SAVED_MAX) :: CRLim) /= (#const RLIM_INFINITY) &&
-      other == (#const RLIM_SAVED_MAX) = ResourceLimitUnknown
+    | ((#const RLIM_SAVED_MAX) :: CRLim) /= (#const RLIM_INFINITY)
+    , other == (#const RLIM_SAVED_MAX)
```
hasufell, line 107: So GHC only sees redundant pattern match if you use `,` instead of `&&`?

How did you spot it then?

sheaf: No, the redundant pattern match happens in both cases.

I had changed the code here, as I was using `considerAccessible` on the second pattern match and it was more consistent with `,` rather than `&&`. I didn't change it back, but I can do so if you prefer.

hasufell: I think what is most important is that the unix release on hackage matches what GHC ships with.

hasufell:
> using 'considerAccessible' causes an issue with the Safe Haskell
annotation, as GHC.Exts isn't Safe;

I think it's fine to replace `Safe` with `Trustworthy`.

Bodigrim:
I'm fine either way, disabling `-Woverlapping-patterns` does not bother me.

Bodigrim:
@sheaf any chance you can finish it up this weekend? We'd like to make a release.

sheaf:
I've changed it back to use `considerAccessible`, if you think that's OK.

hasufell (approved):

Bodigrim (approved):

Bodigrim:
@sheaf and the same, but targeting the master branch please.

PR #239, opened 2022-07-25
Use ByteString for GroupEntry/UserEntry

### LINKED ISSUE #237: `GroupEntry`/`UserEntry` uses String (even in ByteString variants)
https://github.com/haskell/unix/issues/63#issuecomment-1179805976

String has encoding issues. These things should be NUL terminated byte arrays that we get from the C API. The problem is how to unify across String and ByteString modules, since these are types, not just functions:

1. use ByteString based record for both APIs
2. keep String based records
3. provide two different types
4. parametrize the type with the String type
5. don't expose the constructor/records, but provide "getters" (these can do the `ByteString -> String` conversion in the String API)

hasufell:
No PosixString instance for now. Not sure
if it's interesting.

Fixes #237 

@Bodigrim

### DIFF
--- a/System/Posix/User/Common.hsc
+++ b/System/Posix/User/Common.hsc
@@ -1,0 +1,79 @@
+{-# LANGUAGE Safe #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  System.Posix.User.Common
+-- Copyright   :  (c) The University of Glasgow 2002
+-- License     :  BSD-style (see the file libraries/base/LICENSE)
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  provisional
+-- Portability :  non-portable (requires POSIX)
+--
+-- POSIX user\/group support
+--
+-----------------------------------------------------------------------------
+
+module System.Posix.User.Common where
+
+import Data.ByteString ( ByteString )
+import System.Posix.Types
+
+#include "HsUnix.h"
+
+
+#if defined(HAVE_PWD_H)
+import Foreign.Ptr
+import Foreign.Marshal
+import Foreign.Storable
+import Data.ByteString ( packCString )
+
+-- internal types
+data {-# CTYPE "struct passwd" #-} CPasswd
+data {-# CTYPE "struct group"  #-} CGroup
+
+data LKUPTYPE = GETONE | GETALL
+
+unpackGroupEntry :: Ptr CGroup -> IO GroupEntry
+unpackGroupEntry ptr = do
+   name    <- (#peek struct group, gr_name) ptr >>= packCString
+   passwd  <- (#peek struct group, gr_passwd) ptr >>= packCString
+   gid     <- (#peek struct group, gr_gid) ptr
+   mem     <- (#peek struct group, gr_mem) ptr
+   members <- peekArray0 nullPtr mem >>= mapM packCString
+   return (GroupEntry name passwd gid members)
+
+unpackUserEntry :: Ptr CPasswd -> IO UserEntry
+unpackUserEntry ptr = do
+   name   <- (#peek struct passwd, pw_name)   ptr >>= packCString
+   passwd <- (#peek struct passwd, pw_passwd) ptr >>= packCString
+   uid    <- (#peek struct passwd, pw_uid)    ptr
+   gid    <- (#peek struct passwd, pw_gid)    ptr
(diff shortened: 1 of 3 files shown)


System/Posix/User.hsc
```diff
@@ -371,13 +377,30 @@ getEffectiveUserName = do
...
+-- | Manually constructing 'GroupEntry' in String modules is discouraged. It will truncate
+-- Chars to 8bit. Use 'System.Posix.User.ByteString' instead.
+pattern GroupEntry :: String          -- ^ The name of this group (gr_name)
+                   -> String          -- ^ The password for this group (gr_passwd)
+                   -> GroupID         -- ^ The unique numeric ID for this group (gr_gid)
+                   -> [String]        -- ^ A list of zero or more usernames that are members (gr_mem)
+                   -> GroupEntry
+pattern GroupEntry gn gp gi gm <- MkGroupEntry (C8.unpack -> gn) (C8.unpack -> gp) gi (fmap C8.unpack -> gm) where
+  GroupEntry gn gp gi gm = MkGroupEntry (C8.pack gn) (C8.pack gp) gi (C8.pack <$> gm)
+#if __GLASGOW_HASKELL__ >= 802
```
Bodigrim, line 401: We do not support GHC 8.0 anyway.

System/Posix/User/Common.hsc
```diff
@@ -0,0 +1,79 @@
...
+   gecos  <- (#peek struct passwd, pw_gecos)  ptr >>= packCString
+#endif
+   dir    <- (#peek struct passwd, pw_dir)    ptr >>= packCString
+   shell  <- (#peek struct passwd, pw_shell)  ptr >>= packCString
+   return (MkUserEntry name passwd uid gid gecos dir shell)
+
+#endif // HAVE_PWD_H
+
+data UserEntry =
+ MkUserEntry {
```
Bodigrim, line 63:
```suggestion
 UserEntry {
```
And re-export it directly from `ByteString.hsc`.

Bodigrim (approved):

Bodigrim:
@hs-viktor @vdukhovny could you please take a look?

vdukhovni (commented):
Mostly LGTM.

My only concern is that the GECOS field may well be UTF8 encoded on some systems.  It is human-edited/readable data and in various non-US locales could well hold strings in that locale.  Of course that locale's LC_CTYPE need not be UTF8.

At the very least we should call out the change.  Users of the `String` API might not get the string they expect or were used to.  They could switch to the ByteString API and apply the relevant encoding.  Or we could consider attempting to apply the locale's character set to the GECOS field, though of course the user's locale may not match the system-administrator's locale (and then various non-Haskell utilities would also show garbage for non-ASCII gecos fields).

Perhaps try to decode in the current locale, and if that fails, just unpack the ByteString?

Or would it be better to just always unpack (current code) and leave the caller to repack and decode in some suitable encoding???

hasufell:
> Mostly LGTM.
> 
> My only concern is that the GECOS field may well be UTF8 encoded on some systems.  It is human-edited/readable data and in various non-US locales could well hold strings in that locale.  Of course that locale's LC_CTYPE need not be UTF8.
> 
> At the very least we should call out the change.  Users of the `String` API might not get the string they expect or were used to.  They could switch to the ByteString API and apply the relevant encoding.  Or we could consider attempting to apply the locale's character set to the GECOS field, though of course the user's locale may not match the system-administrator's locale (and then various non-Haskell utilities would also show garbage for non-ASCII gecos fields).
> 
> Perhaps try to decode in the current locale, and if that fails, just unpack the ByteString?
> 
> Or would it be better to just always unpack (current code) and leave the caller to repack and decode in some suitable encoding???

`getpwnam` is not POSIX. Any manpage I looked at doesn't specify an encoding. It's NUL terminated char pointer.

Whether most systems use UTF-something is irrelevant. Guessing the encoding based on current locale is wrong.

The current string implementation assumes Latin-1 and doesn't do unicode decoding either. This PR maintains that behavior for the String module. I don't think changing that makes it any more or less correct. Both are wrong.

Only the ByteString variant is correct.

vdukhovni:
Since the original (pre-PR) code is equivalent to just an `unpack` of the GECOS, then indeed we're not really changing anything. I had assumed that it previously applied a locale-specific decoder, but that's not the case.

vdukhovni (approved):

PR #245, opened 2022-09-10
Consistently use `throwErrnoPathIf*`

NicolasT:
In some functions, a combination of `throwErrnoIf*` and `modifyIOError`/`ioeSetFileName` was used, even though `throwErrnoPathIf*` helpers are available.

This patch cleans up such code, unifying the use of the `throwErrnoPathIf*` helpers.

PR #247, opened 2022-09-10
Fix WASI build

NicolasT:
The `System.Posix.Files.PosixString` and `System.Posix.Temp.PosixString` modules failed to build on the WASI/WASM platform because foreign imports of `mkstemp` and `mknod` was not conditional on `configure` findings.

This patch aligns these modules with their non-`PosixString` counterparts.

See: https://github.com/haskell/unix/pull/245
See: https://github.com/haskell/unix/pull/246
See: https://github.com/haskell/unix/runs/8284652815?check_suite_focus=true

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image_family: freebsd-13-0
+  image_family: freebsd-13-1
 
 task:
   install_script: pkg install -y ghc hs-cabal-install git autoconf
--- a/System/Posix/Files/PosixString.hsc
+++ b/System/Posix/Files/PosixString.hsc
@@ -107,6 +107,11 @@ import System.Posix.PosixPath.FilePath
 
 import Data.Time.Clock.POSIX (POSIXTime)
 
+#if !defined(HAVE_MKNOD)
+import System.IO.Error ( ioeSetLocation )
+import GHC.IO.Exception ( unsupportedOperation )
+#endif
+
 -- -----------------------------------------------------------------------------
 -- chmod()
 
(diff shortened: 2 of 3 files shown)


NicolasT:
Not sure this will fix all issues, I need to rely on CI to find other missing conditional compilation sections...

While drafting this patch, I did find something I'd suggest could warrant a change/improvement: currently, the code is as follows:

```haskell
#if !defined(HAVE_FOO)
{-# WARNING foo "..." #-}
foo :: A -> B -> IO C
foo _ _ = ioError ...
#else
foo :: A -> B -> IO C
foo a b = ...
#endif // HAVE_FOO
```

This works, but has one drawback: one could, by mistake, write different types for both functions (especially in the `ioError` one, say, `foo :: B -> A -> IO C` or `foo :: A -> B -> IO D` would  compile just fine), at which point the API would be different depending on the `configure` result for `HAVE_FOO`.

Instead, I'd suggest this alternative approach:
```haskell
foo :: A -> B -> IO C
#if !defined(HAVE_FOO)
{-# WARNING foo "..." #-}
foo _ _ = ioError ...
#else
foo a b = ...
#endif // HAVE_FOO
```

NicolasT:
#246 could depend on this one to succeed in CI, or this one could depend on #246 to succeed in CI. I could cherry-pick 452d826a325bb931e76e5eb346cd88c10d506410 into my branch to at least unblock the FreeBSD build for this PR.

NicolasT:
(I now cherry-picked 452d826a325bb931e76e5eb346cd88c10d506410 from #246 into my branch to unblock FreeBSD CI. Could drop that commit if desired.)

NicolasT:
:tada:

hasufell:
@TerrorJack

TerrorJack (approved):
lgtm!

TerrorJack:
> Instead, I'd suggest this alternative approach:

Yes, it's a good way to avoid the type signature mismatch based on configure result, definitely recommended for new code.

Bodigrim (approved):

hasufell (approved):

Bodigrim:
Thanks @NicolasT!

PR #248, opened 2022-09-10
ci: test GHC 9.4(.2)

NicolasT:
Since `base` 4.17 is supported by the library, actually test GHC 9.4 in CI.


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        ghc: ['9.2', '9.0', '8.10', '8.8', '8.6', '8.4', '8.2']
```
Bodigrim, line 17: I suspect you need to specify `9.4.2` to make it work.

NicolasT: Not sure. The issue appears to be GHC 9.4(.2) is already installed in the MacOS environment, and `ghcup` doesn't want to proceed unless `--force` is used. However, since `ghcup` is invoked by `haskell/actions/setup`, I need to figure out how to pass things through.

hasufell: No, this is a bug in ghcup and haskell actions uses an ancient version.

I suggest to drop haskell setup action and just use straight ghcup with a manual matrix.

NicolasT: Is it an option to use HaskellCI (as was done for TravisCI in the past, looking at some leftover files in the repo) to set up all Linux jobs?

hasufell: Sure

NicolasT:
Giving up on this one for now.

PR #249, opened 2022-09-10
ci: update Fedora tests to Fedora 36

NicolasT:
Fedora 34 has been EOL since 2022-06-07.

See: https://docs.fedoraproject.org/en-US/releases/eol/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,10 +60,10 @@ jobs:
         autoreconf -i
         cabal test all --test-show-details=direct
 
-  fedora34:
+  fedora36:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
   fedora34:
```
Bodigrim, line 63: Please rename, either to `fedora36` or just `fedora`.

NicolasT: Ugh, of course. Done.

hasufell (approved):

Bodigrim (approved):

PR #252, opened 2022-10-02
Use capi for syscalls that break under musl's handling of 64-bit time_t

redneb:
I maintain a [repo with builds of GHC](https://github.com/redneb/ghc-alt-libc) for the musl C standard library and I encountered a subtle bug of this library that affects GHC under 32-bit architectures with musl.

In 32-bit architectures, there was a transition where the C type `time_t` was redefined to be 64-bit in order to address the Y2038 problem. The way this was handled by musl was by introducing new versions of all affected syscalls that work with 64-bit `time_t` (e.g. `utimensat_time64` is like `utimensat` but with 64-bit `time_t`). In addition, a redirect was introduced to create an alias of the new versions of these functions under the old name (e.g. [here's the redirection](https://git.musl-libc.org/cgit/musl/tree/include/sys/stat.h?id=dc9285ad1dc19349c407072cc48ba70dab86de45#n119) for `utimensat`).

The problem is that this redirection is defined as a C macro in a C header file and as such, it does not get picked by GHC when the foreign function import is done via `ccall`, but works just fine when `capi` is used. So this PR changes all affected syscalls to be imported with `capi`, which should be a fairly uncontroversial change.

### DIFF
--- a/System/Posix/DynamicLinker/Prim.hsc
+++ b/System/Posix/DynamicLinker/Prim.hsc
@@ -85,10 +85,17 @@ data RTLDFlags
   | RTLD_LOCAL
     deriving (Show, Read)
 
+#if defined(HAVE_DLFCN_H)
 foreign import capi unsafe "dlfcn.h dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
 foreign import capi unsafe "dlfcn.h dlsym"  c_dlsym  :: Ptr () -> CString -> IO (FunPtr a)
 foreign import capi unsafe "dlfcn.h dlerror" c_dlerror :: IO CString
 foreign import capi unsafe "dlfcn.h dlclose" c_dlclose :: (Ptr ()) -> IO CInt
+#else
+foreign import ccall unsafe "dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
+foreign import ccall unsafe "dlsym"  c_dlsym  :: Ptr () -> CString -> IO (FunPtr a)
+foreign import ccall unsafe "dlerror" c_dlerror :: IO CString
+foreign import ccall unsafe "dlclose" c_dlclose :: (Ptr ()) -> IO CInt
+#endif // HAVE_DLFCN_H
 
 packRTLDFlags :: [RTLDFlags] -> CInt
 packRTLDFlags flags = foldl (\ s f -> (packRTLDFlag f) .|. s) 0 flags
--- a/System/Posix/DynamicLinker/Prim.hsc
+++ b/System/Posix/DynamicLinker/Prim.hsc
@@ -1,3 +1,4 @@
+{-# LANGUAGE CApiFFI #-}
 {-# LANGUAGE Trustworthy #-}
 {-# OPTIONS_GHC -Wno-trustworthy-safe #-}
 
@@ -84,10 +85,10 @@ data RTLDFlags
   | RTLD_LOCAL
     deriving (Show, Read)
 
-foreign import ccall unsafe "dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
-foreign import ccall unsafe "dlsym"  c_dlsym  :: Ptr () -> CString -> IO (FunPtr a)
-foreign import ccall unsafe "dlerror" c_dlerror :: IO CString
-foreign import ccall unsafe "dlclose" c_dlclose :: (Ptr ()) -> IO CInt
+foreign import capi unsafe "dlfcn.h dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
+foreign import capi unsafe "dlfcn.h dlsym"  c_dlsym  :: Ptr () -> CString -> IO (FunPtr a)
+foreign import capi unsafe "dlfcn.h dlerror" c_dlerror :: IO CString
+foreign import capi unsafe "dlfcn.h dlclose" c_dlclose :: (Ptr ()) -> IO CInt
 
 packRTLDFlags :: [RTLDFlags] -> CInt
 packRTLDFlags flags = foldl (\ s f -> (packRTLDFlag f) .|. s) 0 flags
--- a/System/Posix/DynamicLinker/Prim.hsc
+++ b/System/Posix/DynamicLinker/Prim.hsc
@@ -86,15 +86,15 @@ data RTLDFlags
     deriving (Show, Read)
 
 #if defined(HAVE_DLFCN_H)
-foreign import capi unsafe "dlfcn.h dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
+foreign import capi safe "dlfcn.h dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
 foreign import capi unsafe "dlfcn.h dlsym"  c_dlsym  :: Ptr () -> CString -> IO (FunPtr a)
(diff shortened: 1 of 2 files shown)


Bodigrim:
@TerrorJack could you please advise on wasm32 job? I remember there were some changes wrt `dlfcn.h`.

Bodigrim (approved):

redneb:
It appears that in wasm32-wasi there is no `dlfcn.h`, but the following line does compile successfully:
```
foreign import ccall unsafe "dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
```
which is what we currently use in this package. So I think the solution here is to add an `#ifdef` in order to use `ccall` for wasm32-wasi (since `ccall` doesn't require a header file to be provided) and use `capi` elsewhere.

I pushed a commit that does that and the CI tests are passing everywhere now.

System/Posix/DynamicLinker/Prim.hsc
```diff
@@ -84,10 +85,17 @@ data RTLDFlags
   | RTLD_LOCAL
     deriving (Show, Read)
 
+#if defined(HAVE_DLFCN_H)
+foreign import capi unsafe "dlfcn.h dlopen" c_dlopen :: CString -> CInt -> IO (Ptr ())
```
vdukhovni, line 89: The `dlopen` call can be comparatively expensive.  One can reasonably argue that this should be `safe` without a significant performance penalty.  Once the object is loaded, the other calls should be quick.

Of course the same would then apply to the `ccall` version.

vdukhovni (approved):
LGTM, modulo perhaps the making `dlopen` be `safe`???

System/Posix/DynamicLinker/Prim.hsc
redneb: I agree that `dlopen` should be `safe` and I would be happy to push another commit to do that. But I think this is part of a wider problem with this package (see #34) and I don't think we should tackle the problem in a piecemeal fashion. So I would prefer to not do this as part of this PR.

vdukhovni: > I agree that `dlopen` should be `safe` and I would be happy to push another commit to do that. But I think this is part of a wider problem with this package (see #34) and I don't think we should tackle the problem in a piecemeal fashion. So I would prefer to not do this as part of this PR.

I can accept that I guess, though I tend to take a more opportunistic approach. Big bang fixes can be difficult, but spot improvements while looking at adjacent code are sometimes a good way to make progress.  I'll leave the judgement call to you and others.

hasufell: I think "this is bad elsewhere too" isn't a good argument against fixing. 

Rather fix it locally here and open a new ticket to fix the rest as well, unless this is controversial.

redneb: @vdukhovni @hasufell I pushed added an additional commit that switches the imports of `dlopen` & `dlclose` to `safe`. The reason that `dlopen` needs `safe` is obvious; it interacts with the local file system. But `dlclose` also has to be `safe` because it might trigger the execution of destructors defined in the library.

hasufell:
How do we know this is the only issue with header redirection? This frequently happens on darwin system too. I'd be fine with making all foreign imports  use `capi`. Any opinions, @vdukhovni ?

vdukhovni:
> How do we know this is the only issue with header redirection? This frequently happens on darwin system too. I'd be fine with making all foreign imports use `capi`. Any opinions, @vdukhovni ?

My take is that whenever possible we should use `capi`.

redneb:
Here's how I came up with the changes included in the PR: I found the [commit](https://git.musl-libc.org/cgit/musl/commit/?id=1febd21d3fb7b78f6fc13e48b3ff2a396a607a15) in the musl repo where the redirect was introduced and examined all changes there. I also searched for the `_REDIR_TIME64` CPP symbol in all C headers. I compared all that against all of GHC's source tree (including the `libraries` directory) and found all the affected places. There were all in 3 libraries: [`directory`](https://github.com/haskell/directory/pull/145), [`time`](https://github.com/haskell/time/pull/225), and here (the other PRs have been merged already).

> I'd be fine with making all foreign imports use capi. Any opinions, @vdukhovni ?

Agreed, `capi` should be the default policy for this repo.

hasufell:
> Agreed, capi should be the default policy for this repo.

So then we can just remove the CPP ifdef and keep capi?

redneb:
> So then we can just remove the CPP ifdef and keep capi?

Unfortunately not. As I explained in a previous post here, it's need for `wasm32-wasi`.

hasufell (approved):

liskin:
I'm wondering whether (and if so, why?) this fix is enough—I believe that the `instance Storable CTimeVal` implementation will still `peek`/`poke` both members (`tv_sec` and `tv_usec`) as 32-bit values, as they're both declared as `CLong`. This might silently work if the memory happens to be zeroed (and https://github.com/xmonad/X11/pull/100#issuecomment-2080418403 suggests it often is in practice) but one thing is certain—CLong is not Y2038-safe on 32-bit platforms.

Can someone please check if my thinking is correct? Thanks!

liskin:
> Can someone please check if my thinking is correct? Thanks!

I asked the Debian folks as well and it appears that `CTimeVal` isn't really being used in practice ([detailed explanation](https://salsa.debian.org/haskell-team/DHG_packages/-/commit/3b08f9950cf7f485ea712530ba4b5ba2a149ac8d#note_485312)). Might not apply to other platforms but Linux/Debian should be all right then.

hasufell:
@liskin 

From the debian discussion 

> My understanding is that we are good with this patch in Debian, but this should definitely be fixed in the unix library if they want to support other systems / OSes.

Can you submit a PR?

liskin:
> From the debian discussion
> 
> > My understanding is that we are good with this patch in Debian, but this should definitely be fixed in the unix library if they want to support other systems / OSes.
> 
> Can you submit a PR?

Suppose I can try but right now I have absolutely no idea what platform to use which would be affected. Turns out Debian i386 isn't adopting 64-bit time_t, only armhf, armel and a few others. And ever there, the code in question isn't being used (as we established earlier). Suppose I could just fix `CTimeVal` in isolation and test in a Debian armhf VM… but it's not exactly the most important thing I could be doing today :-D

hasufell:
We have armv7 CI: https://github.com/haskell/unix/actions/runs/8825026384/job/24228587469

liskin:
> We have armv7 CI: https://github.com/haskell/unix/actions/runs/8825026384/job/24228587469

Yeah, but it'll take months/years before Debian's t64 stuff trickles down to Ubuntu, and even then `CTimeVal` won't be used at all because of some `#ifdef`s. So that CI won't help me, and also that CI won't even start failing in a couple months time because the issue doesn't affect glibc/musl at all because `utimensat` is being used which takes `CTimeSpec` instead.

hasufell:
Even in absence of a reproducer, I am assuming there is a theoretical fix. What is it? Can you provide a PR for it?

liskin:
> Even in absence of a reproducer, I am assuming there is a theoretical fix. What is it? Can you provide a PR for it?

Yes: https://github.com/haskell/unix/pull/318
I mean, I don't _really_ know what I'm doing (as in, not as much as I'd like to) but the CI won't run in my fork (arm needs custom runners and MacOS uses a different major version ans fails) so PR it is.

PR #255, opened 2022-10-27
Drop config.sub from repository

bgamari:
This file is generated by autoreconf, which the user must run
anyways when building from git.


bgamari:
It appears that this has already been done in `master`.

PR #256, opened 2022-11-01
Add semWaitInterruptible

sheaf:
The `semWaitInterruptible` function allows us to start a thread that blocks on the semaphore but can be interrupted.

### DIFF
--- a/System/Posix/Semaphore.hsc
+++ b/System/Posix/Semaphore.hsc
@@ -1,5 +1,6 @@
 {-# LANGUAGE Safe #-}
 {-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE InterruptibleFFI #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  System.Posix.Semaphore
@@ -16,7 +17,7 @@
 
 module System.Posix.Semaphore
     (OpenSemFlags(..), Semaphore(),
-     semOpen, semUnlink, semWait, semTryWait, semThreadWait,
+     semOpen, semUnlink, semWait, semWaitInterruptible, semTryWait, semThreadWait,
      semPost, semGetValue)
     where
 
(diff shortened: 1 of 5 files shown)


sheaf:
[Here's](https://hackage.haskell.org/package/semaphore-compat-1.0.0/candidate) the API I would like to publish on Hackage which makes use of this new function. (I'm waiting on having a version of `unix` before publishing.)

Bodigrim:
@sheaf could you please upload Haddocks for the candidate? 

I'm very much illiterate in this area, so happy to defer the judgement to @hasufell and @hs-viktor.

System/Posix/Semaphore.hsc
```diff
@@ -79,6 +80,20 @@ semWait (Semaphore fptr) = withForeignPtr fptr semWait'
     where semWait' sem = throwErrnoIfMinus1Retry_ "semWait" $
                          sem_wait sem
 
+-- | Lock the semaphore, blocking until it becomes available.
+--
+-- Unlike 'semWait', this wait operation can be interrupted with an
+-- EINTR signal.
```
hasufell, line 86: You mean a `SIGPIPE` signal? `eINTR` is the error code. But that's an implementation detail of the RTS, isn't it? Any `throwTo` can interrupt.

https://downloads.haskell.org/~ghc/9.2.5/docs/html/users_guide/exts/ffi.html#interruptible-foreign-calls

hasufell:
can we add a test to https://github.com/haskell/unix/blob/master/tests/Semaphore001.hs

System/Posix/Semaphore.hsc
vdukhovni: > You mean a `SIGPIPE` signal? `eINTR` is the error code. But that's an implementation detail of the RTS, isn't it? Any `throwTo` can interrupt.
> 
> https://downloads.haskell.org/~ghc/9.2.5/docs/html/users_guide/exts/ffi.html#interruptible-foreign-calls

More precisely, actually interrupted by a `throwTo` from another thread. The use of `SIGPIPE` on Unix systems for this purpose is an implementation detail. And of course `EINTR` is not the signal but rather a feature of the return state.

vdukhovni (approved):
LGTM, modulo haddock wording.

sheaf:
I've added a test, updated the documentation, and [uploaded the candidate haddocks for `semaphore-compat`](https://hackage.haskell.org/package/semaphore-compat-1.0.0/candidate). Let me know if anything else needs to be done. Thanks!:

sheaf:
Looks like I still need to address some issues with the test. I'll let you know when I'm done.

Bodigrim:
> Looks like I still need to address some issues with the test. I'll let you know when I'm done.

Fixed: different test executables have been using the same `semOpen "/test"` blocking each other in unexpected manner.

@hasufell good to go?

Bodigrim (approved):

sheaf:
Fantastic, thanks for seeing this through @Bodigrim, I've been rather swamped lately.

hasufell (approved):

PR #258, opened 2022-11-18
Avoid `Data.List.{head,tail}`

Bodigrim:
CLC has approved the proposal to add `{-# WARNING #-}` to `Data.List.{head,tail}` (https://github.com/haskell/core-libraries-committee/issues/87). It means that usage of `head` and `tail` will emit compile-time warnings.

This patch eliminates the only usage of `head` in `unix`.


Bodigrim:
Closed in favor of https://github.com/haskell/unix/pull/259.

PR #259, opened 2022-11-18
Avoid `Data.List.{head,tail}`

Bodigrim:
CLC has approved the proposal to add `{-# WARNING #-}` to `Data.List.{head,tail}` (https://github.com/haskell/core-libraries-committee/issues/87). It means that usage of `head` and `tail` will emit compile-time warnings.

This patch eliminates the only usage of `head` in `unix`.

### DIFF
--- a/.github/workflows/ci-wasm32-wasi.yml
+++ b/.github/workflows/ci-wasm32-wasi.yml
@@ -12,16 +12,16 @@ jobs:
       - name: setup-ghc-wasm32-wasi
         run: |
           pushd $(mktemp -d)
-          curl -L https://github.com/tweag/ghc-wasm32-wasi/archive/refs/heads/master.tar.gz | tar xz --strip-components=1
+          curl -L https://gitlab.haskell.org/ghc/ghc-wasm-meta/-/archive/master/ghc-wasm-meta-master.tar.gz | tar xz --strip-components=1
           ./setup.sh
-          ~/.ghc-wasm32-wasi/add_to_github_path.sh
+          ~/.ghc-wasm/add_to_github_path.sh
           popd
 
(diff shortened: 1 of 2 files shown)


hasufell (approved):
This changes behaviour though right? Passing in an empty dir wasn't possible before this patch?

Bodigrim:
@hasufell good point. I filed #260 to track the issue with `mdir = Just ""`, but scaled back this PR to preserve the existing (partial) semantics, making the very minimal change just to avoid `head` and `tail`. How does it look now?

hasufell:
Sure, but why does `head` cause a warning and `last` doesn't?

Bodigrim:
The CLC proposal asked to add `{-# WARNING #-}` for `head` and `tail` only, but not for `last` and `init`. The rationale behind that was that it is considerably more difficult to migrate away from `last` and `init`: one cannot easily replace them with pattern matching and there is no `unsnoc :: [a] -> Maybe ([a], a)` available.

hasufell:
> The CLC proposal asked to add `{-# WARNING #-}` for `head` and `tail` only, but not for `last` and `init`. The rationale behind that was that it is considerably more difficult to migrate away from `last` and `init`: one cannot easily replace them with pattern matching and there is no `unsnoc :: [a] -> Maybe ([a], a)` available. 

I wouldn't mind if we inline `unsnoc` internally until it hits base.

https://hackage.haskell.org/package/extra-1.7.12/docs/src/Data.List.Extra.html#unsnoc

Bodigrim:
@hasufell good to go? I replaced `last` with our own `unsnoc`.

hasufell (approved):

Issue #264, opened 2022-12-05
Include GHC 9.4 in CI ?


andreasabel:
I have come here because I get a library problem with `unix` under the `ubuntu-22.04` virtual environment when used with GHC 9.4.2 and static linking:
```
/home/runner/.ghcup/ghc/9.4.2/lib/ghc-9.4.2/lib/../lib/x86_64-linux-ghc-9.4.2/unix-2.7.3/libHSunix-2.7.3.a(Files.o):
function ghczuwrapperZC1ZCunixzm2zi7zi3ZCSystemziPosixziFilesZCmknod: 
error: undefined reference to '__xmknod'
```
(See https://github.com/agda/agda/actions/runs/3603326548/jobs/6071389168#step:11:38.)

Maybe such problems could be caught if the CI here included GHC 9.4 and also something that tests static linking (`--enable-executable-static`).

Is there a reason that GHC 9.4 isn't included in the CI?  After all, it ships with `unix`.

hasufell:
> Is there a reason that GHC 9.4 isn't included in the CI? 

Mostly that we don't get paid.

Bodigrim:
A PR adding GHC 9.4 to the build matrix would be welcome.

andreasabel:
> A PR adding GHC 9.4 to the build matrix would be welcome.

See:
- #266

PR #266, opened 2022-12-06
CI: include GHC 9.4, bump actions to latest version

andreasabel:
Re: 
- #264

CI: include GHC 9.4, bump actions to latest version (where possible).
~~Had to stick to ubuntu-20.04 as e.g. 8.2.2 does not install on 22.04.~~

SQUASH me!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macOS-latest]
+        os: [ubuntu-22.04, macOS-latest]
         ghc: ['9.4', '9.2', '9.0', '8.10', '8.8', '8.6', '8.4', '8.2']
     steps:
     - uses: actions/checkout@v3
+    - name: Install prerequisites for GHC 8.2 on ubuntu-22.04
+      if: runner.os == 'Linux' && matrix.ghc == '8.2'
+      run: |
+        sudo apt-get install libncurses5 libtinfo5
     - uses: haskell/actions/setup@v2
       id: setup-haskell-cabal
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,7 @@
 name: ci
 on:
-  push:
-    branches: [ master, ci-* ]
-  pull_request:
-    branches: [ master, ci-* ]
+  - push
+  - pull_request
 
 defaults:
   run:
@@ -13,7 +11,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         os: [ubuntu-22.04, macOS-latest]
         ghc: ['9.4', '9.2', '9.0', '8.10', '8.8', '8.6', '8.4', '8.2']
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,17 +13,17 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
hasufell, line 16: Why?

.github/workflows/ci.yml
```diff
@@ -11,17 +13,17 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, macOS-latest]
-        ghc: ['9.2', '9.0', '8.10', '8.8', '8.6', '8.4', '8.2']
+        os: [ubuntu-20.04, macOS-latest]
```
hasufell, line 18: Why not latest?

.github/workflows/ci.yml
andreasabel: Because with `fail-fast: true`, if one fails, you do not get information about the status of the other matrix entries.
(It's like the old GHC, which could only give you _one_ error.)

.github/workflows/ci.yml
andreasabel: GHC 8.2.2 has no install on `ubuntu-22.04`.  
See: https://github.com/andreasabel/unix/actions/runs/3624403077/jobs/6111349578
> Error: All install methods for ghc 8.2.2 failed

hasufell: It does, the haskell setup action just swallows the error: https://github.com/andreasabel/unix/actions/runs/3624403077/jobs/6111349578#step:3:50

The most common cause is different libncurses.

Try installing `libncurses5` and `libtinfo5`.

andreasabel: > Try installing `libncurses5` and `libtinfo5`.

That helped, thanks!

Bumped matrix to `ubuntu-22.04` then.

Bodigrim (commented):
Thanks, much appreciated!

.github/workflows/ci.yml
```diff
@@ -1,7 +1,9 @@
 name: ci
 on:
-  - push
-  - pull_request
+  push:
+    branches: [ master, ci-* ]
+  pull_request:
+    branches: [ master, ci-* ]
```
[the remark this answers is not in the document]
andreasabel: The problem with not limiting to `master` is that in the standard workflow (push to a branch, open a PR) _every workflow is duplicated_, also on each subsequent push to the same PR.

I think the namespace for branches should be partitioned into these (1) where you want CI to always run (I usually choose above two variants) and those (2) where CI is only run on PRs.  So, if you want CI but don't want to open a PR, you choose a branch name `ci-XXX`, and if you want to open a PR, some other branch name.

.github/workflows/ci.yml
andreasabel: Sure, if GHA resources are a problem, but then I think the duplication of CI runs in PRs should also be addressed.

.github/workflows/ci.yml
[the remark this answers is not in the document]
Bodigrim: Current maintainers rarely push feature branches directly to `haskell/unix`. Instead all work and CI jobs happen in forks, and PRs are raised once ready.

I don't mind to discuss other work conventions, assigning special meaning to `ci-*` branches, but let's not scope creep.

.github/workflows/ci.yml
andreasabel: Ok.

.github/workflows/ci.yml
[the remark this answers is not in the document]
andreasabel: Fine, let's postpone this, and hope that Saint Greta will got get us.

Bodigrim (approved):
Thanks!

Bodigrim:
@hasufell is it good to go? (Squash, not merge)

hasufell (approved):

andreasabel:
> @hasufell is it good to go? (Squash, not merge)

Yes, from my side it is done.

PR #268, opened 2022-12-13
Add missing autoconf checks for chown/fchdir/fchmod

TerrorJack:
These are not present in wasm32-wasi, and is needed for fixing GHC #22589.


hasufell (approved):

Bodigrim (approved):

vdukhovni (approved):
This assumes that having `chown(2)` implies also having `fchown(2)`, which is probably reasonable in practice, but a more precise test could also check for `fchown(2)`.
I'm approving anyway, in the hope that the assumption will hold on all relevant platforms.

PR #270, opened 2022-12-14
Ditch haskell setup action

hasufell:


hasufell:
This should be ready now.

.github/workflows/ci.yml
```diff
@@ -21,19 +21,26 @@ jobs:
...
-      with:
-        ghc-version: ${{ matrix.ghc }}
+    - if: runner.os == 'Linux' # this is actually a bug in github images: https://github.com/actions/runner-images/pull/6770
+      name: Fix ghcup dir
+      run: echo GHCUP_INSTALL_BASE_PREFIX=/usr/local >> $GITHUB_ENV
+    - name: Setup toolchain
+      run: |
+        which ghcup
+        ghcup install cabal recommended
+        ghcup install ghc --set ${{ matrix.ghc }}
```
andreasabel, line 31: Would be nice if `ghcup` could also install the prerequisites or print a command that could be executed to do so.

As it stands, you might have to continue adding steps to install prerequisites to various ghc versions --- but you might not be scared of that.

andreasabel (commented):
I guess the idea behind `haskell/actions/setup` is to install GHC with its prerequisites.
I also thought of ditching this action, since it is currently badly maintained, but in general such an action makes sense to have and use.  
Of course if `ghcup` had a `sudo` mode also installing prerequisites, then it would fully replace the setup action.
I am neutral on this PR, but please go ahead if you think it is the right step to do.

Bodigrim:
Let's wait for dust to settle (e. g., https://github.com/actions/runner-images/pull/6770) before accumulating more workarounds.

Ideally I'd prefer to keep `haskell/actions/setup`, I just don't get why it cannot get job done reliably.

.github/workflows/ci.yml
hasufell: > Would be nice if ghcup could also install the prerequisites

It will never do so.

hasufell: > As it stands, you might have to continue adding steps to install prerequisites to various ghc versions --- but you might not be scared of that.

Yes, I'm fine with that. The action doesn't add any value and instead makes it hard to reason about what exactly is installed and from what source.

hasufell:
> Ideally I'd prefer to keep haskell/actions/setup, I just don't get why it cannot get job done reliably.

It is unreliable and over-engineered and it's hard to reason about what bindist it installs and from which source.

There's no benefit.

hasufell:
The dust has settled. I want to merge this soon.

Bodigrim:
I would generally prefer haskell/action, but as long as you are willing to maintain the barebone job, I’m fine with it.

Bodigrim (approved):

PR #275, opened 2023-03-17
Fix segfault for `System.Posix.Env.ByteString.getEnvironment`

### LINKED ISSUE #274: `System.Posix.Env.ByteString.getEnvironment` segfaults on empty environment
Repro:
```haskell
import System.Posix.Env.ByteString

main :: IO ()
main = do
  clearEnv
  getEnvironment >>= print
```
The the technical issue is a missing null pointer check, but the real reason is that we have a lot of copy & paste code that has diverged.

sol:
(fixes #274)

### DIFF
--- a/System/Posix/Env.hsc
+++ b/System/Posix/Env.hsc
@@ -28,15 +28,15 @@ module System.Posix.Env (
 
 #include "HsUnix.h"
 
+import Foreign hiding (void)
 import Foreign.C.Error (throwErrnoIfMinus1_)
 import Foreign.C.Types
 import Foreign.C.String
-import Foreign.Marshal.Array
-import Foreign.Ptr
-import Foreign.Storable
 import Control.Monad
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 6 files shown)


sol:
@hasufell apparently `master` is broken (see #276).

hasufell:
I'm not sure master is broken. This could be a permission issue with your fork. There are cases where CI works on the main repo, but not when initiated via a fork.

hasufell:
Ok, I reran the job on master and it also fails on NetBSD: https://cirrus-ci.com/build/4510365560602624

I believe @Bodigrim added it

sol:
Regarding the other test failures:
```
unix-tests: printenv: executeFile: does not exist (No such file or directory)
```
It's not entirely clear to me why they don't fail on `master`.  Any ideas?

sol:
Actually, I do know. I need to protect the environment for that test I added.

sol:
Ok, CI (mostly) passes.  From my side this is good to go.

PR #279, opened 2023-03-25
Ensure that FilePaths don't contain interior NULs

hasufell:
Follows:

* https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10110
* https://github.com/haskell/core-libraries-committee/issues/144

----

@bgamari @Bodigrim


System/Posix/ByteString/FilePath.hsc
```diff
@@ -131,3 +139,23 @@ throwErrnoPathIfMinus1_  = throwErrnoPathIf_ (== -1)
 throwErrnoTwoPathsIfMinus1_ :: (Eq a, Num a) => String -> RawFilePath -> RawFilePath -> IO a -> IO ()
 throwErrnoTwoPathsIfMinus1_  loc path1 path2 =
     throwErrnoIfMinus1_ (loc <> " '" <> BC.unpack path1 <> "' to '" <> BC.unpack path2 <> "'")
+
+-- | Check an encoded 'FilePath' for internal NUL octets as these are
+-- disallowed in POSIX filepaths. See #13660.
```
Bodigrim, line 144:
```suggestion
-- disallowed in POSIX filepaths. See https://gitlab.haskell.org/ghc/ghc/-/issues/13660.
```
And same below.

Bodigrim:
Overall looks good. @vdukhovni could you please take a look?

System/Posix/PosixPath/FilePath.hsc
```diff
@@ -42,17 +43,20 @@ import System.OsPath.Types
...
 #if !MIN_VERSION_base(4, 11, 0)
 import Data.Monoid ((<>))
 #endif
 
 
 withFilePath :: PosixPath -> (CString -> IO a) -> IO a
-withFilePath = useAsCString . getPosixString
+withFilePath path f = do
+  checkForInteriorNuls path
+  useAsCString (getPosixString path) f
```
vdukhovni, line 59: Likewise here, use `useAsCStringLen`

vdukhovni (changes requested):
As mentioned in upthread comments, I think that just using `useCStringLen` is simpler and faster.

System/Posix/ByteString/FilePath.hsc
```diff
@@ -50,7 +56,9 @@ import Data.Monoid ((<>))
 type RawFilePath = ByteString
 
 withFilePath :: RawFilePath -> (CString -> IO a) -> IO a
-withFilePath = useAsCString
+withFilePath path f = do
+  checkForInteriorNuls path
+  useAsCString path f
```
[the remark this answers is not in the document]
hasufell: Yes, @Bodigrim already suggested that in the GHC MR, I simply forgot to move it here

System/Posix/ByteString/FilePath.hsc
```diff
@@ -50,7 +57,9 @@ import Data.Monoid ((<>))
 type RawFilePath = ByteString
 
 withFilePath :: RawFilePath -> (CString -> IO a) -> IO a
-withFilePath = useAsCString
+withFilePath path f = do
+  checkForInteriorNuls path
```
vdukhovni, line 61: Why are we converting the path to a CString twice, optimising for the failure case.  Why not assume success and use the `ptr` directly if the length check passes???

vdukhovni (changes requested):

System/Posix/ByteString/FilePath.hsc
hasufell: Makes sense

Bodigrim (approved):

System/Posix/ByteString/FilePath.hsc
```diff
@@ -131,3 +138,27 @@ throwErrnoPathIfMinus1_  = throwErrnoPathIf_ (== -1)
...
+        else ioError err
+  where
+    err =
+        IOError
+          { ioe_handle = Nothing
+          , ioe_type = InvalidArgument
+          , ioe_location = "checkForInteriorNuls"
+          , ioe_description = "POSIX filepaths must not contain internal NUL octets."
+          , ioe_errno = Nothing
+          , ioe_filename = Just (either (error . show) id
```
vdukhovni, line 158: Under what conditions are we hiding a bottom (error . show) inside ioe_filename?
I don't think that examining the error should ever blow up...

vdukhovni (changes requested):

System/Posix/ByteString/FilePath.hsc
hasufell: We are not. This function is total due to https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-IO-Encoding-Failure.html#v:TransliterateCodingFailure

This is even tested: https://github.com/haskell/filepath/blob/98f8bba9eac8c7183143d290d319be7df76c258b/tests/abstract-filepath/EncodingSpec.hs#L102

vdukhovni: > We are not. This function is total due to https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-IO-Encoding-Failure.html#v:TransliterateCodingFailure
> 
> This is even tested: https://github.com/haskell/filepath/blob/98f8bba9eac8c7183143d290d319be7df76c258b/tests/abstract-filepath/EncodingSpec.hs#L102

Then I think it would be more clear to return `Nothing` for the ioe_filename if unexpectedly that function fails anyway, and *otherwise* return `Just` the decoded filename.
```
either (const Nothing) Just :: Either a b -> Maybe b
```

hasufell: I disagree. Using Nothing seems to potentially mask a theoretical bug in the encoders or base library. Under the hood we use unsafePerformIO and certain low level IO primitives.

I want the code to crash, but I don't think it's worthwhile to move it into IO either.

vdukhovni: Examining fields of a thrown exception should never throw another error, no matter how unlikely that might be.
Another alternative might be to simply render the filename octets as characters with no attempt to interpret as UTF8. Then the whole issue goes away.  It is not uncommon for filenames to be in some 8bit charset that isn't UTF8.

hasufell: I definitely do not want to mask a potential bug with Nothing. So the only other option is to move it into IO.

However, we already use the `error` pattern elsewhere:

https://github.com/haskell/unix/blob/dc91c94e6231ca7ac9ae0214956b30089b6b0def/System/Posix/PosixPath/FilePath.hsc#L139

I don't see the value of complicating all this code.

Changing the way we print filepaths in errors is not a trivial change. I'm aware examining error messages programmatically is not a good thing, but it's commonly used (e.g. in test suites), so this might potentially break stuff.

vdukhovni: That `_toStr` example is also wrong IMNSHO.  When we have an octet string with no hope of knowing whether it is some encoded text string, or just 8-bit data (sans NUL octets), or when it is a text string, what encoding produced it, the best we can do is render exactly one character per octet, without any "decoding".

Note also that in that context `Nothing` is not an option, as it is with `ioe_filename`.  It is not OK to ever booby-trap the IOError object fields with latent bottoms.

hasufell: > That _toStr example is also wrong IMNSHO. When we have an octet string with no hope of knowing whether it is some encoded text string, or just 8-bit data (sans NUL octets), or when it is a text string, what encoding produced it, the best we can do is render exactly one character per octet, without any "decoding".

There's nothing wrong about this, as these are user visible error messages, not proper filepaths.

I'm well aware of all the encoding intricacies of filepaths, since I've implemented the AFPP.

This is a tradeoff choice.

hasufell: Another angle that I'm willing to discuss is whether `ioe_filename` should allow the caller to recover the actual filename or whether it's just descriptive.

In that context, converting every Word8 to a Char is very specific semantics the caller would have to know about. I'd go so far to say it's wrong on this level, since most other stuff in base uses `getFileSystemEncoding`. So we'd use [decodeFS](https://hackage.haskell.org/package/filepath-1.4.100.3/docs/System-OsPath-Posix.html#v:decodeFS) (which has correctness issues too, btw).

So, I don't think there is any right choice if we want the caller to be able to recover the filename. 

If we just want nice printing, the current solution is ok.

hasufell: > It is not OK to ever booby-trap the IOError object fields with latent bottoms.

I agree and our tests could trivially catch this. As I pointed out, these functions are total and `error` is unreachable.

hasufell: So, Bytestring modules do the infamous unpack:

https://github.com/haskell/unix/blob/dc91c94e6231ca7ac9ae0214956b30089b6b0def/System/Posix/ByteString/FilePath.hsc#L133

I'm rather negative to this approach and would argue the fact that this exists is due to lack of `decodeFS` for bytestring (which isn't an issue in OsPath modules).

However, one could make the consistency argument.

hasufell:
We haven't come to a conclusion wrt `ioe_filename` and I was actually planning to consult with GHC devs about this. Or even CLC. It'a not clear what the API of this record is to the end user.

Based on that, we can more easily decide what to do.

vdukhovni:
I don't feel very strongly about encoding choices, but I am quite adamant about error values not containing bottoms.
Pick the encoding strategy that will be least surprising, and never ever have any of the fields of IOError throw when forced.

When working with paths that are ByteStrings, if they make sense as UTF8, assume they're UTF8, otherwise raw?
Assuming the default system filename encoding (as suggested) is another.  No encoding choice will be always right, we simply don't know.  Part of the reason the IO operation failed, could well be a poor choice of encoding... :-(

hasufell:
> but I am quite adamant about error values not containing bottoms

Yes, as I explained above, the current code using `TransliterateCodingFailure` never throws.

If we pick another encoding, we have to document it in base, IMO: https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-IO-Exception.html#v:ioe_filename

It'll be utterly confusing to end users otherwise, looking for the properties of this record field.

hasufell:
Correct. I need to update the PR.

hasufell:
Should be settled.

vdukhovni (approved):

PR #284, opened 2023-06-26
Bump base bound to <4.20

bgamari:
For GHC 9.8.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -31,12 +31,12 @@ task:
     image: family/pg-ci-netbsd-vanilla-9-3
     platform: netbsd
   install_script:
-    - export PKG_PATH="http://cdn.NetBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f '1 2' -d.)/All/"
-    - pkg_add ghc cabal-install git autoconf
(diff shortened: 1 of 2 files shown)


bgamari:
Changes validated; feel free to merge when ready.

hasufell (approved):

hasufell:
@Bodigrim text is causing problems again on NetBSD:

```
Failed to build text-2.0.1. The failure occurred during the configure step.
Build log (
/tmp/.cabal/logs/ghc-9.2.1/text-2.0.1-dd9b99f9d9265ef039a60ddb10373037a4d55295184cd64795cba80e4cb5b990.log
):
Configuring library for text-2.0.1..
cabal: Missing dependencies on foreign libraries:
* Missing (or bad) C libraries: c++, c++abi
This problem can usually be solved by installing the system packages that
provide these libraries (you may need the "-dev" versions). If the libraries
are already installed but in a non-standard location then you can use the
flags --extra-include-dirs= and --extra-lib-dirs= to specify where they are.If
the library files do exist, it may contain errors that are caught by the C
compiler at the preprocessing stage. In this case you can re-run configure
with the verbosity flag -v3 to see the error messages.

cabal: Failed to build text-2.0.1 (which is required by unix-2.8.1.0). See the
build log above for details.
```

I tried to disable `simdutf` flag, but it still fails: https://cirrus-ci.com/task/5089109375451136

This C++ stuff has been quite a nuisance, unfortunately.

hasufell:
@Bodigrim it does not work: https://github.com/haskell/unix/pull/284/checks?check_run_id=14717521228

hasufell:
I'm disabling NetBSD for now. I don't have time or interest to debug this.

PR #286, opened 2023-07-08
Allow bytestring-0.12

clyring:
`unix` does not use any of the functions that have changed behavior in `bytestring-0.12.0.0`, so the bound-bump is safe. (See the [changelog](https://hackage.haskell.org/package/bytestring-0.12.0.0/changelog).)

I've also replaced the one use of the now-deprecated `Data.ByteString.Internal.memcpy` with the equivalent `Foreign.Marshal.Utils.copyBytes`.

### DIFF
--- a/System/Posix/Env/ByteString.hsc
+++ b/System/Posix/Env/ByteString.hsc
@@ -42,7 +42,7 @@ import System.Posix.Env ( clearEnv )
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Char8 as BC
 import Data.ByteString (ByteString)
-import Data.ByteString.Internal (ByteString (PS), memcpy)
+import Data.ByteString.Internal (ByteString (PS))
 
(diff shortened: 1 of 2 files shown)


System/Posix/Env/ByteString.hsc
```diff
@@ -133,7 +133,7 @@ putEnv (PS fp o l) = withForeignPtr fp $ \p -> do
   --
   -- hence we must not free the buffer
   buf <- mallocBytes (l+1)
-  memcpy buf (p `plusPtr` o) l
+  copyBytes buf (p `plusPtr` o) l
```
hasufell, line 136: I can confirm that `copyBytes` exists since at least [base-4.9.1.0](https://hackage.haskell.org/package/base-4.9.1.0/docs/Foreign-Marshal-Utils.html#v:copyBytes).

hasufell (approved):

PR #288, opened 2023-07-17
Make error strings consistent

hasufell:
System.Posix.ByteString.FilePath just unpacks, while System.Posix.PosixPath.FilePath tried to be smart and decode with UTF (without failing).

This divergence was discussed here:

* https://github.com/haskell/unix/pull/279
* https://github.com/haskell/core-libraries-committee/issues/189

----

Although the discussion at the CLC tracker hasn't come to a conclusion yet, I'm setting this up as the favorite approach so far. I suggest that we also augment the documentation in base.

@vdukhovni @Bodigrim


hasufell:
Aband

PR #285, opened 2023-07-04
JS: add support for utimes/lutimes/futimes

hsyl20:
In particular, add support for `touchFd` required here: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9237

Cc @bgamari

PR #290, opened 2023-07-19
Fix building and testing with the JS backend

hsyl20:
Note that most tests have been disabled because they don't pass due to missing C function support (e.g. pipe). But it's a good basis to start adding tests for supported things (e.g. touchFd, cf #285).

This requires a recent version of GHC that provides `getMonotonicTimeNSec`. See
https://gitlab.haskell.org/ghc/ghc/-/issues/23687 and https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10900.

### DIFF
--- a/System/Posix/Signals.hsc
+++ b/System/Posix/Signals.hsc
@@ -112,7 +112,17 @@ import System.Posix.Process
 import System.Posix.Process.Internals
 import Data.Dynamic
 
+#if defined(javascript_HOST_ARCH)
+-- These signals aren't supported by the JS RTS
+-- and the header isn't exposed.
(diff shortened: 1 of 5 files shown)


hsyl20:
For some reason, Fedora mirror site is 404

bgamari:
Thanks for doing this, @hsyl20 !

bgamari (approved):

hsyl20:
Yes otherwise the platform isn't recognized by the configure script.

hsyl20:
(we overwrite these files when we build the package in the ghc tree)

hasufell (approved):

PR #294, opened 2023-09-12
Fix `withFilePath`

hasufell:
https://github.com/haskell/cabal/pull/9241

@Bodigrim we'll need another release. This bug was introduced by 2.8.2.0, so maybe we can mark it as deprecated on hackage?

### DIFF
--- a/System/Posix/ByteString/FilePath.hsc
+++ b/System/Posix/ByteString/FilePath.hsc
@@ -153,7 +153,8 @@ decodeWithBasePosix ba = B.useAsCStringLen ba $ \fp -> peekFilePathPosix fp
 -- | Wrapper around 'useAsCString', checking the encoded 'FilePath' for internal NUL octets as these are
 -- disallowed in POSIX filepaths. See https://gitlab.haskell.org/ghc/ghc/-/issues/13660
 useAsCStringSafe :: RawFilePath -> (CString -> IO a) -> IO a
-useAsCStringSafe path f = useAsCStringLen path $ \(ptr, len) -> do
+useAsCStringSafe path f = useAsCString path $ \ptr -> do
+    let len = B.length path
     clen <- c_strlen ptr
     if clen == fromIntegral len
         then f ptr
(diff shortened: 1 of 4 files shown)


hasufell:
@bgamari

vdukhovni (approved):

vdukhovni:
Sorry about that.  I think the fault was mostly mine.  Mind you the docs of `useAsCStringLen` could be more clear. They don't mention lack of NUL-termination, one has to carefully contrast with `useAsCString` to notice that only the latter mentions NUL-termination, and so implicitly perhaps the former does not NUL-terminate. I guess that choice is *natural*...

hasufell:
Well, I was well aware of this intricacy, because I worked long enough on the ShortByteString code. So I'm not sure how that slipped through my radar either.

Bodigrim:
Thanks for a quick turnaround!

I did not follow the discussion closely, is it possible to add a regression test?

In the meantime I've revised [`unix-2.8.2.0` with `base < 0`](https://hackage.haskell.org/package/unix-2.8.2.0/revisions/).

hasufell:
Yes, we could generate random filepaths, run `withFilePath` and just print the CString.

I don't know if I will have time to do that today.

Bodigrim:
A simple unit test will do IMHO, just to prevent someone in future to refactor back to `useAsCStringLen`.

hasufell:
> A simple unit test will do IMHO, just to prevent someone in future to refactor back to `useAsCStringLen`.

I don't think so. This is a memory bug and doesn't always trigger.

bgamari:
Thanks for the heads up! Given the severity of the issue and the fact that it made it into a release, it would probably be a good idea to open a ticket and point to it from the changelog. I have done for former in #295; @hasufell, perhaps you can do the latter?

erikd:
I just got hit by this. :tada:

Bodigrim:
@hasufell could you please rebase?

Bodigrim (approved):

hasufell:
rebased

hasufell:
CI all green. Changelog updated. Version bumped to 2.8.2.1

PR #299, opened 2023-09-26
Don't `foreign import` `environ`

bgamari:
`musl` defines the `environ` symbol as a weak alias, which on AArch64 requires particular treatment by the compiler. As GHC doesn't have access to the prototype of the symbol, it doesn't know that this treatment is necessary and consequently we emit assembly that the linker will choke on.

Fixes [GHC #24011](https://gitlab.haskell.org/ghc/ghc/-/issues/24011).

### DIFF
--- a/System/Posix/Env/Internal.hsc
+++ b/System/Posix/Env/Internal.hsc
@@ -22,8 +22,11 @@ getCEnviron = nsGetEnviron >>= peek
 foreign import ccall unsafe "_NSGetEnviron"
    nsGetEnviron :: IO (Ptr (Ptr CString))
 #else
-getCEnviron = peek c_environ_p
+getCEnviron = _getCEnviron
 
-foreign import ccall unsafe "&environ"
-   c_environ_p :: Ptr (Ptr CString)
+-- N.B. we cannot import `environ` directly in Haskell as it may be a weak symbol
+-- which requires special treatment by the compiler, which GHC is not equipped to
+-- provide. See GHC #24011.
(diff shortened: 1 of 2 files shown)


System/Posix/Env/Internal.hsc
```diff
@@ -22,8 +22,11 @@ getCEnviron = nsGetEnviron >>= peek
 foreign import ccall unsafe "_NSGetEnviron"
    nsGetEnviron :: IO (Ptr (Ptr CString))
 #else
-getCEnviron = peek c_environ_p
+getCEnviron = _getCEnviron
 
-foreign import ccall unsafe "&environ"
```
hasufell, line 27: Does `capi` not work the same?

Bodigrim (approved):

System/Posix/Env/Internal.hsc
bgamari: As far as I can tell, no. `capi` can only be used to import functions and things that behave like functions.

hasufell (approved):

bgamari:
@hasufell, would it be possible to cut a minor release with this fix? This will otherwise prevent us from offering AArch64/musl bindists for GHC 9.8.1.

hasufell:
> @hasufell, would it be possible to cut a minor release with this fix? This will otherwise prevent us from offering AArch64/musl bindists for GHC 9.8.1.

Can

bgamari:
Do note that for us to ship the new release it will need to be available by Friday.

PR #311, opened 2024-01-23
PosixString.hsc: expose Fd

domenkozar:


domenkozar:
Quite a few of functions have `Fd` in their type signature, so now I need to import it from `System.IO` if I want to use them.

The way I see it, it already exposes the API on the `Fd` so it just makes it convenient to have it as one import.

The same goes for `SeekMode`.

hasufell:
I don't have a strong opinion, but it should be consistent.

hasufell:
ping

hasufell:
Is there any interest in pursuing this further?

PR #313, opened 2024-02-03
Use self-hosted FreeBSD runners

hasufell:


.cirrus.yml
```diff
@@ -1,13 +1,3 @@
-task:
-  name: FreeBSD
-  freebsd_instance:
-    image_family: freebsd-13-1
-  install_script: pkg install -y ghc hs-cabal-install git autoconf
-  script:
-    - cabal update
-    - autoreconf -i
-    - cabal test --test-show-details=direct
```
Bodigrim, line 9: Let's keep it, maybe just comment it out, in case the self-hosted runner goes down. Who owns it at the moment?

hasufell: Me

Bodigrim (approved):

PR #318, opened 2024-04-30
Fix CTimeVal definition for platforms where time_t isn't CLong

liskin:
One such platform is Debian unstable armhf, which is in the process of transitioning to 64-bit time_t: https://wiki.debian.org/ReleaseGoals/64bit-time

On that platform (as well as any other glibc/musl platform), however, CTimeVal isn't used for anything at all because there are #ifdefs that prefer using `utimensat` which takes CTimeSpec instead. This fix is therefore quite theoretical, as it is unknown whether there are any platforms actually affected.

Related: https://github.com/haskell/unix/pull/252

### DIFF
--- a/System/Posix/Files/Common.hsc
+++ b/System/Posix/Files/Common.hsc
@@ -551,7 +551,7 @@ foreign import capi unsafe "sys/stat.h futimens"
     c_futimens :: CInt -> Ptr CTimeSpec -> IO CInt
 #endif
 
-data CTimeVal = CTimeVal CLong CLong
+data CTimeVal = CTimeVal (#type time_t) (#type suseconds_t)
 
 instance Storable CTimeVal where
     sizeOf    _ = #size struct timeval


Bodigrim:
@liskin could you please rebase? Hopefully CI will get better.

liskin:
> @liskin could you please rebase? Hopefully CI will get better.

Done, pls reapprove the CI workflows.

hasufell:
Posix standard is very clear: https://pubs.opengroup.org/onlinepubs/009604599/basedefs/sys/time.h.html

```
time_t         tv_sec      Seconds. 
suseconds_t    tv_usec     Microseconds.
```

So this doesn't seem theoretical to me at all. It must be fixed.

liskin:
> So this doesn't seem theoretical to me at all. It must be fixed.

Yeah, sure, it's just a bit silly that we have no idea whether there is an actual platform where that code is being used, and if so, what that platform is…

hasufell:
@hs-viktor any opinions?

Bodigrim:
@vdukhovni what do you think about this?

hasufell:
I'm fairly confident about this.

@liskin ping me in a week (or at Zurihac) and I'll merge, unless there are objections in the meantime.

liskin:
> @liskin ping me in a week (or at Zurihac) and I'll merge, unless there are objections in the meantime.

Won't be attending ZuriHac this year unfortunately, so just a ping. 🍻

Bodigrim:
Thanks!

PR #319, opened 2024-04-30
Drop old MacOS jobs

Bodigrim:
Drop old MacOS jobs: there are no bindists of old GHCs for `aarch64`, which `macos-latest` now refers to.

We could potentially stick to `macos-13` instead, but whatever, I don't really care that much about old GHCs.


Bodigrim:
Oh well, there is no preinstalled `ghcup` on `macos-latest` runners. Shall we use `haskell-actions/setup`?

hasufell:
I'm not a fan of the action. We can just do:

```
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_INSTALL_NO_STACK=1 BOOTSTRAP_HASKELL_GHC_VERSION=${{ matrix.ghc }} BOOTSTRAP_HASKELL_ADJUST_BASHRC=yes sh
. ~/.ghcup/env
# install/build
```

hasufell (approved):

PR #324, opened 2024-06-25
Remove `includes` from `.cabal`-file

sol:
This is useless at best, and a common source of confusion.

https://github.com/haskell/cabal/pull/10145
https://github.com/sol/hpack/issues/355

### DIFF
--- a/unix.cabal
+++ b/unix.cabal
@@ -155,9 +155,6 @@ library
     ghc-options: -Wall
 
     include-dirs: include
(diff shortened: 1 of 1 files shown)


Bodigrim (approved):

hasufell:
Not really useless. It is just wrong. These files need not be unconditionally included in every cc invocation.

hasufell (approved):

PR #333, opened 2025-01-01
Fix support for GHCJS_BROWSER compilation flag of GHC

Swordlash:
See https://gitlab.haskell.org/ghc/ghc/-/issues/25613#note_602248

When compiling `ghc` under GHCJS_BROWSER, all the environment checking code should be removed by the preprocessor. This is not the case for the `jsbits` of the `unix` library, which causes troubles for Google Closure Compiler and bundlers. This PR adds missing directives.

### DIFF
--- a/jsbits/time.js
+++ b/jsbits/time.js
@@ -1,4 +1,9 @@
+//#OPTIONS: CPP
+
 function h$js_futimes(fd,atime,mtime) {
+#ifdef GHCJS_BROWSER
+    throw "h$js_futimes unsupported";
+#else
   if (!h$isNode()) {
     throw "h$js_futimes unsupported";
   }
@@ -9,9 +14,13 @@ function h$js_futimes(fd,atime,mtime) {
     return -1;
   }
   return 0;
+#endif
 }
 
 function h$js_utimes(path,path_offset,atime,mtime) {
+#ifdef GHCJS_BROWSER
+    throw "h$js_utimes unsupported";
+#else
   if (!h$isNode()) {
     throw "h$js_utimes unsupported";
   }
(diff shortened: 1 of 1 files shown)


jsbits/time.js
```diff
@@ -1,4 +1,9 @@
+//#OPTIONS: CPP
+
 function h$js_futimes(fd,atime,mtime) {
+#ifdef GHCJS_BROWSER
+    throw "h$js_futimes unsupported";
+#else
   if (!h$isNode()) {
```
Bodigrim, line 7: Could it be 
```suggestion
  if (!h$isNode() || defined(GHCJS_BROWSER)) {
```
?

Bodigrim:
Please rebase cleanly on `master`, ignoring whatever commits were accumulated in GHC mirror.

jsbits/time.js
Swordlash: Not sure what you mean, we want a preprocessor-level `if`, not a runtime-level `if` here.

Bodigrim: Fair enough.

Swordlash: In other words if ghc is compiled with `-DGHCJS_BROWSER` then `h$isNode()` is undefined at runtime and we don't want to include this code at all.

hasufell:
ping @Swordlash

hsyl20:
LGTM.

@luite Could you review it too please?

Swordlash:
Thanks @hasufell :)

Swordlash:
I'm going to close this PR for now since we agreed to remove the macro altogether.

Issue #347, opened 2025-07-02
System.Posix.Directory.Internals.dirEntType always returns DT_UNKNOWN


sjshuck:
Most details can be found here: https://github.com/snoyberg/conduit/issues/459#issuecomment-3029444175

My environment:
* Arch Linux on x86_64
* Stackage Nightly 2025-07-01
* unix-2.8.6.0
* GHC 9.10.2
* cabal 3.14.2.0
* stack 3.7.1

PR #348, opened 2025-07-03
Fix struct dirent d_type macro test

sjshuck:
Fixes https://github.com/haskell/unix/issues/347.  Credit to Andrew Gunnerson (@chenxiaolong).

Test program:

```haskell
module Main (main) where

import Control.Exception                (bracket)
import Data.Function                    (fix)
import Foreign.C.String                 (peekCString)
import System.Posix.Directory
import System.Posix.Directory.Internals
import Text.Printf                      (printf)

printDirEnt :: DirEnt -> IO ()
printDirEnt dirEnt = do
    dName <- dirEntName dirEnt >>= peekCString
    dType <- dirEntType dirEnt
    printf "%-20s%s\n" dName (show dType)

main :: IO ()
main = bracket (openDirStream "/home/steve/foo") closeDirStream $ \dirStream ->
    fix $ \go -> readDirStreamWith printDirEnt dirStream >>= mapM_ (\_ -> go)
```

Output for unix-2.8.7.0:

```
.                   DirType 0
..                  DirType 0
a regular file.txt  DirType 0
a directory.d       DirType 0
a symlink           DirType 0
```

Output with this change:

```
.                   DirType 4
..                  DirType 4
a regular file.txt  DirType 8
a directory.d       DirType 4
a symlink           DirType 10
```


Bodigrim:
@sjshuck could you add a test please?

sjshuck:
I'm a little stuck on how to skip the test if `d_type` isn't defined or if the filesystem doesn't set non-zero values for it.  Open to suggestions.

Bodigrim:
> Do all test containers have a `cc` installed on them?

I think so, these are generic CI machines, not bare minimum containers.

Is `#ifdef HAVE_STRUCT_DIRENT_D_TYPE` inaccessible in the test suite?

sjshuck:
> Is `#ifdef HAVE_STRUCT_DIRENT_D_TYPE` inaccessible in the test suite?

I'm not sure TBH.  I don't know how standard of a macro that is, how much it has to do with GNU, Linux, ...  I did the test another way, compiling optimistically with `d_type`, and if the compilation fails, I'm assuming it's because that isn't defined, and the Haskell test is skipped.  I could scrape stderr and find mention of "d_type" in it to be extra sure, I suppose.

tests/DirEnt.c
```diff
@@ -0,0 +1,49 @@
...
+#include <sys/types.h>
+
+void check_error(const char *msg) {
+    if (errno) {
+        perror(msg);
+        exit(1);
+    }
+}
+
+int main() {
```
hasufell, line 15: I don't understand the purpose of this. We check for headers and library functions in `configure.ac`

https://github.com/haskell/unix/blob/eced23b356ed00510a7e6735b3f093f0c7f94081/configure.ac#L30-L66

sjshuck: Explaining in a comment in the code.

tests/DirEnt.c
```diff
@@ -0,0 +1,73 @@
...
+ * exit with status 0
+ *   This means the "." entry in the DIR stream opened at ".", guaranteed to be
+ *   a directory, has a d_type of DT_DIR.  We should proceed with a test for it
+ *   in Haskell.
+ * exit with status 1
+ *   This means something unexpected went wrong.  Fail the Haskell test also.
+ * exit with stauts 2
+ *   This means the "." entry has a d_type of DT_UNKNOWN.  This is valid; no
+ *   filesystem or operating system is required to yield a useful d_type.
+ *   We should skip testing for non-DT_UNKNOWN values in Haskell.
```
hasufell, line 22: Why can't we do this check in Haskell?

tests/DirEnt.hs
```diff
@@ -0,0 +1,58 @@
...
+
+onFailure :: IO ExitCode -> (ExitCode -> IO ()) -> IO ()
+action `onFailure` after = action >>= \case
+    ExitSuccess -> return ()
+    ec          -> after ec
+infixr 9 `onFailure`
+
+prepareTest :: IO ()
+prepareTest = do
+    system_x "cc --version" `onFailure` exitWith
```
hasufell, line 23: I think this is highly non-portable. I'd prefer to remove the c-tests altogether, do "static" checks in `configure.ac` and do additional runtime checks (whether the kernel just returns `DT_UNKNOWN`) in the Haskell code itself.

If we really want to run C code, we can compile it properly through cabal and use the C FFI to run the appropriate function. I do not understand why we compile manually in the test suite and produce a binary here.

hasufell: > When d_type is defined and the filesystem supports non-zero values for it,

Yes, these absolutely can be tested for in `configure.ac`. I think one of my comments got swallowed by GitHub.

* https://autotools.info/autoconf/custom-tests.html

Bodigrim: > Assume that for Linux, GitHub CI and Hackage CI always make `.` have `d_type` `DT_DIR`, and test exactly that with `dirEntType`, and only run it under Linux. If it returns `UnknownType` or some other value, fail the test noisily, running `mount(8)`, and deal with that when that happens.

I think it's good enough and strikes a good balance. Although I do not get what `mount(8)` reference is about.

hasufell: It's technically a non-portable test and so I think it should be guarded behind a cabal flag. Maybe `non-portable-tests`?

sjshuck: @hasufell I disabled it for non-Linux in `unix.cabal` but maybe that's not what you're talking about.  I set "Allow edits by maintainers" so please feel free to add any Cabal flags.

sjshuck:
Anything else needed of me?

Bodigrim:
@hasufell just a gentle reminder.

sergv:
@hasufell It would be pretty useful to merge this as benefits will be seen by most path traversing libraries. Since path traversing is pretty common operation so number of end users that would benefit is considerable.

hasufell:
Does this require a new release?

sjshuck:
It looks like the diff from v2.8.7.0 is just 1) this non-PVP change and 2) some bounds.  Unless you didn't want to issue a new release for every little change, it'd be nice to get a tiny version (v2.8.7.1).

hasufell:
@Bodigrim I don't think there's much else to wait for. The other PR still needs some discussion imo.

PR #355, opened 2025-09-20
Release 2.8.8.0

hasufell:


sjshuck:
Looks good.   I see you changed the test from `/proc/self` to `.`.  Whether that has a nonzero `d_type` may be dependent on filesystem but for anything likely to be used in CI, it'll be supported, I think.

hasufell:
We test the `d_type` during configure in the C code. If it's 0, the test is skipped.

sjshuck:
Right.  My point is that `/proc/self` is assured to not be skipped on Linux, even if the workspace filesystem of the CI container changes, which is why I chose that.

If you want to use `.` though, that's fine.

hasufell:
What do you mean with "skipped"?

hasufell:
I find this a bit too clever and it relies on non-portable specifities. What we can do is to have a cabal flag to make the test strict: we do not allow skipping it. Since we control the CI, that flag is safe in CI, but will not be enabled on the end users system by default.

Bodigrim:
Tests or no tests, I'd suggest we make a release with the fixed typo soonish, so that there is a chance to get it included into upcoming GHC 9.12.3 and 9.14.1.

hasufell:
I adjusted with a cabal flag.

PR #358, opened 2025-11-29
bytestring >=0.11 bound

### LINKED ISSUE #357: Dependency issues with bytestring
`unix-2.7.3` as available on hackage, declares that it build-depends on `bytestring >=0.9.2`, but it fails to build against `bytestring-0.9.2.1` (on e.g. GHC 8.2.2):
```
System/Posix/Env/ByteString.hsc:152:30: error:                                                                                          
    • Couldn't match expected type ‘CSize’ with actual type ‘Int’                                                                       
    • In the third argument of ‘memcpy’, namely ‘l’                                                                                     
      In a stmt of a 'do' block: memcpy buf (p `plusPtr` o) l                                                                           
      In the expression:                                                                                                                
        do buf <- mallocBytes (l + 1)                                                                                                                       /rows                                                           
           memcpy buf (p `plusPtr` o) l                                                                                                 
           pokeByteOff buf l (0 :: Word8)                                                                                                     cus cells/cols
           throwErrnoIfMinus1_ "putenv" (c_putenv (castPtr buf))                                                                        
    |                                                                                                                                   
152 |   memcpy buf (p `plusPtr` o) l                                                                                                    
    |                              ^                                                                                                         o
Error: [Cabal-7125]                                                                                                                         f
Failed to build unix-2.7.3.                
```

The reason is that `Data.ByteString.Internal.memcpy` changed types:
- `bytestring-0.9.2.1`: `memcpy :: Ptr Word8 -> Ptr Word8 -> CSize -> IO ()`
- `bytestring-0.10.0.0`: `memcpy :: Ptr Word8 -> Ptr Word8 -> Int -> IO ()`

A hackage revision is needed here.

Notably, newer versions of `unix` also declare dependency on `bytestring >=0.9.2`, all the way up to git head, although this issue cannot be reproduced for coincidental reasons: `unix >2.7.3` requires `filepath >=1.4.100`, which requires `bytestring >=0.11.3`.

mniip:
Closes #357.

Prior to `bytestring-0.9.3`, `memcpy` had a different type, breaking `unix-2.7.3` (#357).

Furthermore, since `unix-2.8.0.0` we depend on `filepath >=1.4.100.0`, which already implies `bytestring >=0.11.3`.

### DIFF
--- a/unix.cabal
+++ b/unix.cabal
@@ -83,7 +83,7 @@ library
 
     build-depends:
         base        >= 4.12.0.0  && < 4.23,
-        bytestring  >= 0.9.2     && < 0.13,
+        bytestring  >= 0.11      && < 0.13,
         time        >= 1.9.1     && < 1.16
 
     if flag(os-string)


Bodigrim (approved):

Bodigrim:
Thanks!

PR #360, opened 2025-12-30
Fix OpenBSD CI job

Bodigrim:


unix.cabal
```diff
@@ -194,6 +194,10 @@ test-suite unix-tests
       build-depends: filepath >= 1.4.100.0 && < 1.5.0.0
     ghc-options: -Wall -with-rtsopts=-V0
 
+    -- https://github.com/haskellari/splitmix/issues/101
+    if os(openbsd)
+      build-depends: splitmix < 0.1.3 || > 0.1.3.1
```
hasufell: It appears this is an indirect dependency? If so it shouldn't be in the cabal file, but in cabal.project or something

Bodigrim: Moved it directly into the CI job.

hasufell (approved):

PR #363, opened 2026-01-22
Dynamic CI and extended filestatus fixes

### LINKED ISSUE #364: `fileExtendedStatus` test fails on ARM64 GHC 9.2.8
https://github.com/haskell/unix/actions/runs/21245637397/job/61134120433?pr=363#step:7:637

```
    fileExtendedStatus:          FAIL
      tests/FileExtendedStatus.hs:97:
      expected: ((33024,2049,0,1,1001,1001,0,Just 0,1769079222.758095344s,1769079222.758095344s,1769079222.758095344s),(False,False,False,True,False,False,False))
       but got: ((33024,2049,0,1,1001,1001,0,Just 0,1769079222.758095344s,0s,0s),(False,False,False,True,False,False,False))
      Use -p '/fileExtendedStatus/' to rerun this test only.
```

This is likely due to the fact that we do not check `statx.stx_mask` against the macros that were requested by the user (such as `STX_MTIME`).

This is also described in the manpage:

```
       Note that, in general, the kernel does not reject values in mask
       other than the above.  (For an exception, see EINVAL in errors.)
       Instead, it simply informs the caller which values are supported
       by this kernel and filesystem via the statx.stx_mask field.
       Therefore, do not simply set mask to UINT_MAX (all bits set), as
       one or more bits may, in the future, be used to specify an
       extension to the buffer.
```

hasufell:


.github/workflows/ci.yml
Kleidukos, line 1: One problem with this solution is that you effectively desynchronise the CI matrix and your cabal file's `tested-with` field: 
Screenshot 2026-01-22 at 15-07-33 @hackage › unix — Flora pm


I would recommend that you use https://github.com/Kleidukos/get-tested, whether the Action or its CLI tool directly if you're allergic to actions.

System/Posix/Files/Common.hsc
```diff
@@ -991,91 +992,179 @@ pattern StatxAll = StatxMask 0
...
+-- check for kernel/filesystem support via 'supportsStatxMask'.
+--
+-- E.g.:
+--
+-- @
+-- statx <- getExtendedFileStatus Nothing "unix.cabal" defaultStatxFlags (defaultStatxMask .&. StatxMtime)
+-- if supportsStatxMask statx StatxMtime
+-- then pure (modificationTimeHiResX statx)
+-- else fail "StatxMtime not supported!"
+-- @
```
Kleidukos:
```suggestion
-- > statx <- getExtendedFileStatus Nothing "unix.cabal" defaultStatxFlags (defaultStatxMask .&. StatxMtime)
-- > if supportsStatxMask statx StatxMtime
-- > then pure (modificationTimeHiResX statx)
-- > else fail "StatxMtime not supported!"
```

.github/workflows/ci.yml
hasufell: That requires me to update the `tested-with` field, which I do not want (or only do for releases).

Kleidukos: Yes, so that the field represents reality, so that it can be used by all tooling (Hackage, Flora, CI). You centralise the source of truth instead of having two parallel systems.

Ultimately you do what you want, just be careful to not forget to edit `tested-with` at release. :)

hasufell: Yes, I will synchronize it when I upload it to hackage, not constantly edit it in git and during PRs.

System/Posix/Files/Common.hsc
hasufell: What is the difference?

Kleidukos: Main difference is that with `@`, symbols are interpreted as markup, whereas text on the right of bird tracks (`>`) is interpreted literally. For this specific usecase it does not matter, but using `/` as a file separator or within an operator will be interpreted as the italics/emphasis syntax in Haddock markup when you use `@`, but not when you use `>`. 

(It's a PITA I know)

System/Posix/Files/Common.hsc
```diff
@@ -857,7 +858,7 @@ defaultStatxFlags = mempty
 --
 -- See the pattern synonyms for possible masks. These are combined via @(<>)@.
 -- Masks can be tested via `(.&.)`.
-newtype StatxMask = StatxMask CInt deriving (Read, Show, Eq, Ord, Integral, Num, Enum, Bits, Real)
+newtype StatxMask = StatxMask CUInt deriving (Read, Show, Eq, Ord, Integral, Num, Enum, Bits, Real)
```
Bodigrim: Annoyingly this `CInt` has already been exposed: https://hackage-content.haskell.org/package/unix-2.8.8.0/docs/System-Posix-Files-ByteString.html#t:StatxMask
Could we keep it `CInt` to avoid a major version bump?

hasufell: That would require using unsafe functions like `fromIntegral` now.

Bodigrim: We can rename the constructor and provide a pattern synonym for backward compatibility.

```haskell
newtype StatxMask = StatxMaskUnsigned CUInt ...

pattern StatxMask :: CInt -> StatxMask
pattern StatxMask x = StatxMaskUnsigned (fromIntegral x)
```

hasufell: Had to do

```hs
pattern StatxMask :: CInt -> StatxMask
pattern StatxMask x <- StatxMaskUnsigned (fromIntegral -> x) where
  StatxMask x = StatxMaskUnsigned (fromIntegral x)
```

Because the above actually gives a parse error (functions seem to not be allowed this way). The syntax around pattern synonyms is really horrible.

System/Posix/Files/Common.hsc
Bodigrim:
```suggestion
    StatxMask(.., pattern StatxMask),
```
Let's make it a bundled pattern synonym, so that whenever a use imports `StatxMask(..)` he gets `pattern StatxMask` in scope. This provides for better backward compatibility.

System/Posix/Files/Common.hsc
```diff
@@ -73,7 +74,7 @@ module System.Posix.Files.Common (
     pattern ForceSync,
     pattern DontSync,
     defaultStatxFlags,
-    StatxMask(..),
+    StatxMask(.., pattern StatxMask),
```
Bodigrim:
```suggestion
    StatxMask(.., StatxMask),
```
Sorry, I should test syntax before suggesting.

Bodigrim (approved):

